PR #1, opened 2020-05-05
[WIP] update field name : Event.Event is now Event.Parsed

buixor:


parsers/s02-enrich/crowdsecurity/whitelists.yaml
```diff
@@ -3,11 +3,12 @@ description: "Whitelist events from private ipv4 addresses"
 whitelist:
   reason: "private ipv4 ranges"
 #  ip: 
-#    - 195.154.37.115
+#    - 1.2.3.4
   cidr:
     - "192.168.0.0/16"
     - "10.0.0.0/8"
     - "172.16.0.0/12"
+    - "127.0.0.1"
```
erenJag, line 11: it's missing the CIDR, does it work ?

scenarios/crowdsecurity/http-bf-wordpress_bf.yaml
```diff
@@ -3,7 +3,7 @@ name: crowdsecurity/http-bf-wordpress_bf
 description: "detect wordpress bruteforce"
```
erenJag, line 3: Is it possible to rename this scenario, there is "bf" twice in the name.

parsers/s02-enrich/crowdsecurity/whitelists.yaml
buixor: fixed and moved to IP

scenarios/crowdsecurity/http-bf-wordpress_bf.yaml
buixor: mh not really, I like to keep `<proto>-<attack-type>-<scenario name>` for now

PR #3, opened 2020-05-12
add ftp collection

erenJag:


collections/crowdsecurity/ftp.md
```diff
@@ -0,0 +1 @@
+A collection to defend FTP against bruteforce attack.
```
buixor, line 1: I think in the collection we need to say what logs we support, and give a bit more info about the bruteforce (ie. speed / capacity ?)

buixor: especially if it's **specific** to VSFTPD we need to tell it

parsers/s01-parse/crowdsecurity/ftp-logs.yaml
```diff
@@ -0,0 +1,21 @@
+onsuccess: next_stage
+name: ftp-logs
+description: "Parse FTP (vsftpd) logs"
+filter: "evt.Parsed.program == 'ftp'"
+#debug: true
+pattern_syntax:
+  FTP_AUTH_FAIL: '%{HTTPDERROR_DATE:timestamp} \[pid %{NUMBER}\] \[%{GREEDYDATA:user}\] FAIL LOGIN: Client "%{IP:source_ip}"'
```
buixor, line 7: if we only support the FTP_FAIL_AUTH message, I guess it's worth saying as well :)

scenarios/crowdsecurity/ftp-bf.yaml
```diff
@@ -0,0 +1,13 @@
+type: leaky
+#debug: true
+name: crowdsecurity/ftp-bf
+description: "Detect FTP bruteforce"
+filter: evt.Meta.log_type == 'ftp_failed_auth'
+leakspeed: "10s"
+capacity: 3
```
buixor, line 7: I think "3" is a bit too aggressive, maybe >= 5 ?

buixor:
one more thing :) we need to add it to the collections.json file no ?

erenJag:
Which file are you talking about  ?
Because there is no `collections.json` file, Maybe you talk about the folder collections and yes it contains the yaml file that includes both parser and scenario.

PR #4, opened 2020-05-12
Add modsecurity collections

AlteredCoder:
- Modsecurity parser
- Modsecurity scenario
- modsecurity collection


collections/crowdsecurity/modsecurity.yaml
```diff
@@ -0,0 +1,12 @@
+parsers:
+#generic post-parsing of http stuff
+  - crowdsecurity/modsecurity
+collections:
+  - crowdsecurity/modsecurity
+description: "modsecurity support : modsecurity parser and scenario"
+author: crowdsecurity
+tags:
+  - linux
+  - apache2
```
erenJag, line 10: waf instead of apache ? or add new tag.

collections/crowdsecurity/modsecurity.yaml
```diff
@@ -0,0 +1,12 @@
+parsers:
+#generic post-parsing of http stuff
```
buixor, line 2: I think we wanna remove this comment

collections/crowdsecurity/modsecurity.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - crowdsecurity/modsecurity
+collections:
+  - crowdsecurity/modsecurity
+description: "modsecurity support : modsecurity parser and scenario"
+author: crowdsecurity
+tags:
+  - linux
+  - apache2
+  - web
+  - injection
```
buixor, line 12: do we want to add a `waf` tag rather than injection ?

parsers/s01-parse/crowdsecurity/modsec.yaml
```diff
@@ -0,0 +1,15 @@
+onsuccess: next_stage
+filter: evt.Parsed.program == 'modsecurity'
```
buixor, line 2: were you able to see on a live system if it logs with its own program name ? or it's a supposition ?

PR #11, opened 2020-06-01
Detect access attempts to sensitive data

erenJag:
:warning: This PR depends on https://github.com/crowdsecurity/sec-lists/pull/1/files


scenarios/crowdsecurity/http-sensitive-files.yaml
```diff
@@ -0,0 +1,17 @@
...
+format: 2.0
+debug: true
+name: crowdsecurity/http-sensitive-files
+description: "Detect attempt to access to sensitive files (.log, .db ..) or folders (.git)"
+filter: 'evt.Meta.service == "http" and any(File("sensitive_data.txt"), { evt.Parsed.request endsWith #})'
+groupby: "evt.Meta.source_ip"
+data:
+   - source_url: https://raw.githubusercontent.com/crowdsecurity/sec-lists/master/web/sensitive_data.txt
+   - dest_file: sensitive_data.txt
+capacity: 2
```
buixor, line 11: I think we want to distinct on the URI to limit the impact of false positives :)

scenarios/crowdsecurity/http-sensitive-files.yaml
```diff
@@ -0,0 +1,17 @@
+type: leaky
+format: 2.0
+debug: true
+name: crowdsecurity/http-sensitive-files
+description: "Detect attempt to access to sensitive files (.log, .db ..) or folders (.git)"
+filter: 'evt.Meta.service == "http" and any(File("sensitive_data.txt"), { evt.Parsed.request endsWith #})'
```
buixor, line 6: as well I think the filter is too lose, we should check that it's access or error log no ?

erenJag: done :+1:

scenarios/crowdsecurity/http-sensitive-files.yaml
erenJag: umm, I not sure, if we add `distinct` and an attacker try to access at least 3 different (files and directories in the list) it's just no ?

erenJag: After reflection, you're right.

PR #13, opened 2020-06-01
Scenario: Path traversal probing

AlteredCoder:


scenarios/crowdsecurity/http-path-traversal-probing.yaml
```diff
@@ -0,0 +1,20 @@
+# path traversal probing
+type: leaky
+#debug: true
+name: crowdsecurity/http-path-traversal-probing
+description: "Detect path traversal attempt"
+filter: evt.Meta.service == 'http' and RegexpInFile(evt.Meta.http_path, "http_path_traversal.txt")
```
buixor, line 6: I think we want to restrict the scenario to http acess / error logs

PR #39, opened 2020-10-23
reduce false positives for http-crawl-non_statics

buixor:
- Add `.map` to static extensions in `http-logs` parser
 - Reduce http-crawl-non_statics false-positives by doing `distinct` only on the filename rather than path+query


registergoofy (approved):
lgtm

PR #49, opened 2020-11-13
Maxmind disclaimer

registergoofy:
disclaimer was already there, but not as the exact form it should have. I think they really want to see their website url appear in clear.


buixor (approved):
lgtm

PR #91, opened 2020-12-15
add geoip-enrich parser and telnet-bf scenario

registergoofy:
* crowdsecurity/geoip-enrich
* crowdsecurity.telnet-bf


buixor (approved):
lgtm

PR #96, opened 2020-12-17
seo-bots-whitelists tests

buixor:


postoverflows/s01-whitelist/crowdsecurity/seo-bots-whitelist.yaml
```diff
@@ -1,5 +1,6 @@
 name: crowdsecurity/seo-bots-whitelist
 description: "Whitelist good search engine crawlers"
+debug: true
```
registergoofy, line 3: Is this intended ?

registergoofy (commented):
Is the debug intended ? Ths has fired a new version of the item...

postoverflows/s01-whitelist/crowdsecurity/seo-bots-whitelist.yaml
buixor: oh you are damn right. by the way, isn't something we might want to detect / prevent ?

registergoofy (approved):
lgtm

registergoofy (approved):
LGTM

PR #97, opened 2020-12-17
iptables collection test

buixor:
perform a collection test with :
 - syslog and iptables parser
 - dateparse enricher
 - iptables portscan scenarios


registergoofy (approved):
lgtm

PR #120, opened 2021-01-06
Add external bouncers

erenJag:
* add externals bouncers in the list
* support repositories without releases
* support repositories with prereleases
* add force push in CI to permit the CI to update automatically blockers.json


buixor:
can we look into fetching the social preview image from repositories to handle the image in the hub bouncers list ?

erenJag:
@buixor unfortunately, i was not able to fetch the social media image from the [go-github](https://pkg.go.dev/github.com/google/go-github/v33/github) lib we're using. Moreover, it seems the social media preview is only used when the repository is linked in a social network and the image is quite big (it's a preview).

AlteredCoder (approved):

PR #121, opened 2021-01-06
fix apache user agent parsed field

### LINKED ISSUE #118: apache2 vs nginx : unify output fields of http user agent
The user agent is not stored in the same field in apache2 and nginx :
 - Parsed.http_user_agent in nginx
 - Parsed.agent in apache2

AlteredCoder:
Fix #118


registergoofy (approved):
lgtm

PR #122, opened 2021-01-06
Fix wordpress bf scenario

### LINKED ISSUE #114: scenarios/http-bf-wordpress_bf doesn't work
This scenarios atm doesn't do anything for me.
I have to remove && evt.Parsed.status == '200' or change it to evt.Meta.http_status == '200'.

I think the status is not needed and can ignore some threat.
Some of my server need apache auth to access wp-login, this mean the status is 401 and not 200.
And if some bots try to reach wp-login on non wordpress site this will give à 301 or 404 and then ignore the bot.

AlteredCoder:
Don't check the if status code is 200 but instead if  the HTTP verb is `POST`.

This will also fix #114


buixor (approved):
lgtm

PR #136, opened 2021-01-20
Geoip

registergoofy:
split linux collection in two collection: linux+geoip

### DIFF
--- a/.index.json
+++ b/.index.json
@@ -49,6 +49,24 @@
     "crowdsecurity/http-backdoors-attempts"
    ]
   },
+  "crowdsecurity/geoip": {
+   "path": "collections/crowdsecurity/geoip.yaml",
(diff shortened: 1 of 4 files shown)


erenJag (commented):
If the collection contains only a parser (geoip) there is no need to create a collection. I thought the idea was to create new collection (crowdsecurity/linux-no-geoip) which will not contains geoip parser.

registergoofy:
> If the collection contains only a parser (geoip) there is no need to create a collection. I thought the idea was to create new collection (crowdsecurity/linux-no-geoip) which will not contains geoip parser.

You are right a new collection is created. But the idea was to manage geoip as a collection with wizard.sh because as of now wizard only deal with collection.

registergoofy:
deprecated

PR #155, opened 2021-02-19
Fix seo bots whitelist : deal with empty lines

buixor:
cf. https://github.com/crowdsecurity/crowdsec/issues/630

An empty line in `rdns_*` files would lead to any item validating the expression. Fix this by skipping empty lines.


registergoofy (approved):
lgtm

PR #172, opened 2021-03-04
gh actions fix

he2ss:
* trigger tests when there is a push or pull_request on master
* stop triggers on generate taxonomy when it's pull_request on master


registergoofy (approved):
lgtm

PR #176, opened 2021-03-07
allow laxer apache2 format - fix #157

### LINKED ISSUE #157: apache2 : parser is too restrictive
Currently, the parser for apache2 requires referrer and user-agent to work. However, as reported on gitter, some users with other setups (ie. bitnami), will *not* have referer and user-agents in the logs.

Proposition would be to make those two fields optional :

`      pattern: '(%{IPORHOST:target_fqdn} )?%{COMMONAPACHELOG}( %{QS:referrer} %{QS:http_user_agent})?'`

Instead of :

`      pattern: '(%{IPORHOST:target_fqdn} )?%{COMMONAPACHELOG} %{QS:referrer} %{QS:http_user_agent}'`

buixor:


registergoofy (approved):
lgtm

PR #181, opened 2021-03-12
remove broken scenario `ban-report-ssh_bf_report`

buixor:
as per https://github.com/crowdsecurity/crowdsec/issues/680 this scenario is broken. If a user (as there) tries to install *all* the scenarios of the hub, it's going to break. :grimacing:


registergoofy (approved):
lgtm

PR #187, opened 2021-04-02
haproxy support

buixor:
- haproxy http log parser


parsers/s01-parse/crowdsecurity/haproxy-logs.yaml
```diff
@@ -0,0 +1,23 @@
+filter: "evt.Parsed.program startsWith 'haproxy'"
+onsuccess: next_stage
+name: crowdsecurity/haproxy-logs
+description: "Parse haproxy http logs"
+grok:
+  pattern: '%{HAPROXYHTTPBASE}'
```
blotus, line 6: As we want to move away from patterns distributed with crowdsec itself, maybe create the pattern here directly ?

buixor: yes great idea, fixed

blotus (approved):

AlteredCoder (approved):

Issue #178, opened 2021-03-09
False negative with sshd 


buixor:
As  reported by users, we're missing some sshd-logs, ie.

```
Mar  5 20:24:14 testhost sshd[455340]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=178.200.xxx.xxx  user=root
```

cf. https://discourse.crowdsec.net/t/help-having-only-unparsed-logs/134/4

PR #188, opened 2021-04-02
add grok to catch auth fail log

he2ss:
fixing this https://github.com/crowdsecurity/hub/issues/178


buixor:
LGTM, @he2ss do you know if one failed auth won't generate several lines (including this one that was previously ignored ?)

besides that, looks ok :+1:

buixor (approved):

PR #197, opened 2021-04-21
Improve http bad user agent : use regexp

buixor:
this fixes a false positive where a user coming with a user agent that *contained* a substring that is a bad user agent would be banned : use regepx instead.


registergoofy (approved):
lgtm

PR #199, opened 2021-05-11
Pf merge

buixor:
- Add `pf` collection
 - Add `pf` parser
 - Add `pf` portscan detection


.index.json
```diff
@@ -323,6 +323,47 @@
     "crowdsecurity/http-bf-wordpress_bf"
    ]
   },
+  "firewallservices/lemonldap-ng": {
```
registergoofy, line 326: This should be done by the CI on master, right ?

registergoofy (commented):
.index.json should be updated by the CI ?

.index.json
buixor: yes, re-running the job to avoid this

PR #202, opened 2021-05-19
Pgsql collection

buixor:
- postgres parser
 - postgres bf detection


sabban (approved):
Can you add a unit test?

PR #204, opened 2021-06-02
add express bouncer download from npm api

AlteredCoder:


blockers.go
```diff
@@ -29,6 +36,37 @@ type ItemInfo struct {
...
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return 0, errors.Wrapf(err, "doing request to fetch downloads from NPM API")
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	body, readErr := ioutil.ReadAll(resp.Body)
```
blotus, line 58: Add a check on `resp.Body`,  if it is `nil` it will segfault (maybe return during the previous check as the data we want is in the body anyway ?)

blockers.go
```diff
@@ -29,6 +36,37 @@ type ItemInfo struct {
 	Status      string `json:"status"`
 }
 
+func fetchExpressBouncerDownload() (int, error) {
+	now := time.Now()
+	nowFmt := fmt.Sprintf("%d-%d-%d", now.Year(), now.Month(), now.Day())
+	bouncerFirstReleaseDate := "2021-01-01"
+	url := fmt.Sprintf("https://api.npmjs.org/downloads/point/%s:%s/@crowdsec/express-bouncer", bouncerFirstReleaseDate, nowFmt)
```
blotus, line 43: According to the [documentation](https://github.com/npm/registry/blob/master/docs/download-counts.md#limits) : `All other queries are limited to at most 18 months of data.`
Add a check to make sure we don't request more than 18 months at a time, and if so, split the time range so that we can get all the downloads ?

blotus (approved):

PR #208, opened 2021-06-22
Add support for cpanel

sbs2001:


buixor:
@sbs2001 can we check if the "access logs" created by the cpanel install are parsed by existing apache2 parsers ?

buixor (approved):
LGTM

PR #222, opened 2021-07-23
dovecot is no http

sabban:
dovecot is no http


buixor:
you don't want to add imap/pop3 ?

sabban:
> you don't want to add imap/pop3 ?

We can't know, because it can be both, but one without the other too...

PR #234, opened 2021-08-18
update syslog parser to include datasource type and module

he2ss:
* update syslog parser to include datasource type and module
* update tests

:warning: This MR include new expression that includes objects (Like `evt.Line.Module`) that are only in new crowdsec version that include aquisition revamp.
I think it will fail with old crowdsec version, but as we're versioning hub, it should work properly right ?


buixor:
as the hub is versioned, I don't think it should be an issue

buixor (approved):

AlteredCoder:
Tested, it works :+1:

PR #245, opened 2021-09-13
Wordpress check for bruteforce on xmlrpc

Belphemur:
Also changed the leak speed to be slower, attacker are pacing themselves these days.
I have great success with this to block bad bots.

### DIFF
--- a/collections/crowdsecurity/wordpress.yaml
+++ b/collections/crowdsecurity/wordpress.yaml
@@ -1,6 +1,5 @@
 scenarios:
   - crowdsecurity/http-bf-wordpress_bf
-  - crowdsecurity/http-bf-wordpress_bf_xmlrpc
   - crowdsecurity/http-wordpress_wpconfig
 description: "wordpress: Bruteforce protection and config probing"
 author: crowdsecurity
--- a/collections/crowdsecurity/wordpress.yaml
+++ b/collections/crowdsecurity/wordpress.yaml
@@ -1,7 +1,8 @@
 scenarios:
   - crowdsecurity/http-bf-wordpress_bf
+  - crowdsecurity/http-bf-wordpress_bf_xmlrpc
   - crowdsecurity/http-wordpress_wpconfig
-description: "wordpress : wp-config probing"
+description: "wordpress: Bruteforce protection and config probing"
 author: crowdsecurity
 tags:
   - linux
--- a/scenarios/crowdsecurity/http-bf-wordpress_bf.md
+++ b/scenarios/crowdsecurity/http-bf-wordpress_bf.md
@@ -1,4 +1,4 @@
 Detects bruteforce on wordpress login page 'wp-login.php'.
 
-leakspeed of 2m, capacity of 5
+leakspeed of 10s, capacity of 5
 
--- a/scenarios/crowdsecurity/http-bf-wordpress_bf.md
+++ b/scenarios/crowdsecurity/http-bf-wordpress_bf.md
@@ -1,4 +1,4 @@
 Detects bruteforce on wordpress login page 'wp-login.php'.
 
-leakspeed of 10s, capacity of 5
+leakspeed of 2m, capacity of 5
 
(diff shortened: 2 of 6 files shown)


buixor:
Hello @Belphemur  !

Thanks for the suggestion :+1: 
Do you think it would make sense to have two different scenarios ? (one for wp-login and one for xmlrpc ?)

Thanks

Belphemur:
Hello 👋

I thought about it. But when checking the log I realise the bots are brute forcing both endpoints usually in the same window.

I think it's good for now to have only one scenario.

Sent from Nine
________________________________

he2ss:
Hi @Belphemur,

Thanks for the contribution.
I have just one doubt about the xmlrpc endpoint in wordpress, it's like an API and it always reply with http code 200.
So you can query this API to do other stuff than login, it might generate a lot of false positive for people who use the xmrpc.

Here is an example of bruteforce attack on the xmlrpc : https://nitesculucian.github.io/2019/07/01/exploiting-the-xmlrpc-php-on-all-wordpress-versions/

So maybe we want to separate in two scenarios and add a warning on the xmlrpc one to prevent people who use wordpress API that it can trigger false positive.
Or even better, if you don't use the xmlrpc, you can simply disable it.

Belphemur:
@he2ss I see your point.

I've split the scenario and added it also the the wordpress collection.

Belphemur:
@he2ss & @buixor Any other comments about this ?

Belphemur:
Any decision on this PR ?

he2ss:
Hi @Belphemur,

Sorry for the delay, the issue is that we can't add this scenario to the WordPress collection.
It means that every user using the WordPress collection and have a plugin using xmlrpc will trigger lot of false positive.

The solution will be to remove this scenario from the collection but keep it as a scenario with a warning message in the description that this scenario will raise a lot of false positives.

Belphemur:
Hi @he2ss,

This sound sensible to me. I've removed it from the collection and added the warning.

scenarios/crowdsecurity/http-bf-wordpress_bf.md
```diff
@@ -1,4 +1,4 @@
 Detects bruteforce on wordpress login page 'wp-login.php'.
 
-leakspeed of 10s, capacity of 5
```
he2ss, line 3: You need to keep this scenario with its defaults values.

scenarios/crowdsecurity/http-bf-wordpress_bf.yaml
```diff
@@ -6,9 +6,9 @@ debug: false
 filter: "evt.Meta.log_type == 'http_access-log' && evt.Parsed.file_name == 'wp-login.php' && evt.Parsed.verb == 'POST'"
 groupby: evt.Meta.source_ip
 capacity: 5
-leakspeed: "10s"
```
he2ss, line 9: Same here (no change needed)

he2ss (commented):
Maybe you forgot to cancel those changes.

collections/crowdsecurity/wordpress.yaml
```diff
@@ -1,7 +1,7 @@
 scenarios:
   - crowdsecurity/http-bf-wordpress_bf
   - crowdsecurity/http-wordpress_wpconfig
-description: "wordpress : wp-config probing"
```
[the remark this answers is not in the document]
Belphemur: The description is actually wrong. Since this collection contains the bruteforce protection on wp-login, not just probing config.

scenarios/crowdsecurity/http-bf-wordpress_bf.yaml
Belphemur: reverted in d42f77ca6e0ae9d4d537818eb50d60f72f17c83a

scenarios/crowdsecurity/http-bf-wordpress_bf.md
Belphemur: reverted in d42f77ca6e0ae9d4d537818eb50d60f72f17c83a

he2ss:
@Belphemur Thanks for this contribution!

PR #246, opened 2021-09-16
Slow ssh bf

### LINKED ISSUE #230: creation of a slow brute-force scan of ssh scenario
Hello,
I have a lot of slow sshd scan, and do not match any scenarios. I have modified sshd-bf scenarios with a leakspeed and capacity more hight and it catch better this scans, but all metrics don't participate to the project (because of the alteration os params).
It will be a good idea to create a special scenario tha catch slow scan.
For example I have for 12h :
- 35 different ips with 50 hit
- 30 different ups with 70 hit
- 28 different ips with only one hit

I do some statistics and I see the same IPs every 120s (+ or - 30 sec) 
So a leakspeed with 420sec and a capacity of 6 can detect all of this slow distributed scans

buixor:
fix #230 : add slow ssh bf detection

### DIFF
--- a/.index.json
+++ b/.index.json
@@ -329,15 +329,19 @@
   },
   "crowdsecurity/sshd": {
    "path": "collections/crowdsecurity/sshd.yaml",
-   "version": "0.1",
+   "version": "0.2",
    "versions": {
     "0.1": {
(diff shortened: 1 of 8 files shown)


AlteredCoder (approved):

coh7eiqu8thaBu:
Hello, after more than 48h of tests, the bouncer still have only one connexion to the crowdsec kernel. Good Job !

PR #276, opened 2021-10-27
ModSecurity Event detection for nginx

fstap:
Hi,

I added the possibility to detect ModSecurity events in the nginx logs as the standard ModSecurity parser does not work with nginx. It's a very crude edit as I am not experienced with crowdsec whatsoever. I hope this still helps!


blotus:
Hello @fstap,

Thanks for your PR !

I'm going to close it, we went for a different approach that is more in line with what we did for the modsec apache logs.
You can see the PR here: https://github.com/crowdsecurity/hub/pull/286.

Let us know if you have any issues with the new parser.

PR #292, opened 2021-12-06
add fastly parser and tests

he2ss:


buixor:
@he2ss don't we want to make it a collection instead ?

he2ss:
done

buixor (approved):
lgtm

PR #296, opened 2021-12-08
Traefik parser

AlteredCoder:
Add traefik parser for CLF and JSON format.

Should be compatible with almost all base-http-scenarios


buixor (approved):
lgtm

PR #320, opened 2021-12-16
Fix typo

cronlabspl:
Leaky >> Leak


AlteredCoder:
Hello @cronlabspl,

Thanks for the issue.
This is not a typo but a keyword for the scenario configuration (cf. https://docs.crowdsec.net/docs/scenarios/format/#type)

PR #330, opened 2021-12-29
fix: file extensions for compressed resources

bennet-esyoil:
This adds the .brotli and .gz file extension to the valid resources list.

I think it maybe better to explicitly check for css.gz and css.brotli e.g., but with the `EXT` Regex that would require a change of that. I don't understand the consequences of that, so I'll let the team decide :)

This would close #330


he2ss (approved):
Thanks for your contribution :)

PR #341, opened 2022-01-11
Sync data files with hub index 

sbs2001:
See https://github.com/crowdsecurity/crowdsec/pull/1153


ci.go
```diff
@@ -57,8 +64,13 @@ var types = []string{
 	"scenarios",
 	"postoverflows",
 	"collections",
+	"data_files",
 }
 
+var skipDataFiles bool
+
+// var dataFileConfigs = make([]dataFileConfig, 0)
```
AlteredCoder, line 72: remove commented line please

ci.go
```diff
@@ -89,10 +101,11 @@ func main() {
 	flag.StringVar(&outFile, "output", ".index.json", "File to output index")
 	flag.BoolVar(&generate, "generate", false, "File to output index")
 	flag.StringVar(&inputFile, "input", ".index.json", "File to read index from")
+	flag.BoolVar(&skipDataFiles, "no-data-files", false, "Don't Download data files")
```
AlteredCoder, line 104: Shouldn't we have a target "data_file" like for bouncers or configuration file instead of this flag?

generate.go
```diff
@@ -71,51 +81,62 @@ func (ti *typeInfo) generate(filepath string, configType string) (string, error)
...
+		if fInfo.Description != "" {
+			ti.Description = fInfo.Description
 		}
-		if len(fInfo.Scenarios) > 0 {
-			ti.Scenarios = fInfo.Scenarios
+
+		if fInfo.Labels != nil {
+			ti.Labels = fInfo.Labels
+
+			// var tags_to_keep = []string{"service", "type"}
```
AlteredCoder, line 117: remove commented lines please

generate.go
```diff
@@ -192,3 +221,71 @@ func generateIndex(configType string) (map[string]typeInfo, error) {
...
+
+func getDataFileNameFromPath(p string) string {
+	splits := strings.Split(p, "/")
+	n := len(splits)
+	author := splits[n-2]
+	name := strings.Split(splits[n-1], ".")[0]
+	return author + "/" + name
+}
+
+func fastlyDfProcessor(data []byte) ([]byte, error) {
```
AlteredCoder, line 269: i think we want to have the generation/formating of list in a different binary/code that can be run without the index.json generation

generate.go
sbs2001: This was present before the PR, hence  didn't touch it. Are you sure we should remove this ?

.index.json
```diff
@@ -45,25 +45,9 @@
   },
   "crowdsecurity/base-http-scenarios": {
    "path": "collections/crowdsecurity/base-http-scenarios.yaml",
-   "version": "0.5",
+   "version": "0.1",
```
AlteredCoder, line 48: all the parsers/scenario have been reset to v0.1 ?

PR #364, opened 2022-01-28
Typo

rhoekman:


blotus:
Hello @rhoekman,

Thanks for the PR, but unfortunately, I will have to close it.

Due to the design of the hub, we cannot rename an existing scenario, so we have to live with the typo :(

Issue #366, opened 2022-02-02
crowdsec/http-crawl-non_statics blocking nextcloud


mschiessl:
CrowdSec http-crawl-non_statics recently blocked my nextcloud instance.

Versions:
- crowdsec  v1.3.0
- crowdsecurity/http-crawl-non_statics   enabled  0.2

I fed the nextcloud nginx access log into crowdsec and it seems that CS http-crawl-non_statics scenario ran mad on nextcloud's PROPFIND requests when i was renaming a folder within nextcloud.

Heres the generated (but obfuscated) data:

Alert intel
```
 - ID         :  XXXX
 - Date       : 2022-02-02T03:45:09Z
 - Machine    : obfuscated
 - Simulation : false
 - Reason     : crowdsecurity/http-crawl-non_statics
 - Events Count : 44
 - Scope:Value: Ip: obfuscated
 - Country    : DE
 - AS         : Deutsche Telekom AG
 - Begin      : 2022-02-02 03:45:05.299144812 +0000 UTC
 - End        : 2022-02-02 03:45:06.831214124 +0000 UTC
 ```

Decision
```
  - Active Decisions  :
+---------+-----------------+--------+---------------+----------------------+
|   ID    |   SCOPE:VALUE   | ACTION |  EXPIRATION   |      CREATED AT      |
+---------+-----------------+--------+---------------+----------------------+
| 1303017 | Ip:  obfuscated | ban    | 26m7.4717213s | 2022-02-02T03:45:09Z |
+---------+-----------------+--------+---------------+----------------------+
```

Example:
```
 | datasource_path | /var/www/my-nextcloud/log/nginx-access.log                                   |
| datasource_type | file                                                                          |
| http_args_len   |                                                                             0 |
| http_path       | /remote.php/dav/files/user/dir1/dir2/dir3 |
| http_status     |                                                                           207 |
| http_user_agent | Mozilla/5.0 (Macintosh)                                                       |
|                 | mirall/3.4.1git (build                                                        |
|                 | 8596) (Nextcloud, osx-21.2.0                                                  |
|                 | ClientArchitecture: x86_64                                                    |
|                 | OsArchitecture: x86_64)                                                       |
| http_verb       | PROPFIND  
```
Log Excerpt
![image](https://user-images.githubusercontent.com/5146717/152167438-d0334aaf-c003-46e2-af8e-cdae14e22376.png)

[...]

PR #368, opened 2022-02-03
Http crawl-non-statics nextcloud fix

mschiessl:
This is a fix to solve the issue #366 

-- 
As discussed within the [discord thread](https://discord.com/channels/921520481163673640/938416827237806151),  limiting crawling to GET might be "ok" for now.
Happy to discuss additional methods like "POST / OPTIONS" but most likely they are not a crawling attack and therefore should go into another scenario.


woopstar:
lgtm

buixor:
@mschiessl looks great, thanks. Do you think we want to take "HEAD" into account ? I agree "POST" and "OPTIONS" are likely out of scope.

mschiessl:
Yeah good point - HEAD makes sense in terms of crawling

MikeSchiessl:
I've amended the PR accordingly - hope that makes sense

PR #372, opened 2022-02-03
Fix typo in nextcloud parser and collection doc

umglurf:
fix missing i in identifier in doc, use correct type for syslog/journald


buixor (approved):
lgtm, thanks

PR #386, opened 2022-02-16
Add emby parser, scenario and collection

LePresidente:
Add emby parser, scenario and collection.

Please let me know of any issues


blotus:
Hello @LePresidente,

Thanks for the PR.

The tests are currently failing due to a case mismatch between `lepresidente` and `LePresidente`, once this is fixed we should be able to merge.

PR #390, opened 2022-02-19
Add Ombi parser, scenario and collection

LePresidente:
Add Ombi parser, scenario and collection


he2ss:
Thanks for your contribution @LePresidente.

PR #395, opened 2022-03-01
Add Gitea parser, scenario and collection

LePresidente:
Add Gitea parser, scenario and collection

Works with a single username or Email Address for the username.


parsers/s01-parse/LePresidente/gitea-logs.yaml
```diff
@@ -0,0 +1,26 @@
...
+      statics:
+        - meta: log_type
+          value: gitea_failed_auth
+
+statics:
+    - meta: service
+      value: gitea
+    - meta: user
+      expression: "evt.Parsed.username"
+    - meta: user
```
AlteredCoder, line 21: the `evt.Meta.User` is overwritten by the email address

LePresidente: I think this is right, evt.Meta.User will either be the username or email address in the log line, the scenenario then can be same for both instances

LePresidente: The Grok node uses an if statement so only one will be returned as well (%{EMAILADDRESS:email}|%{USERNAME:username})
if email is empty it will fill username instead

The test passes with this setup currently and the parser.assert shows correctly.

LePresidente: I could change it to greedydata if you would prefer?

AlteredCoder: Hello @LePresidente , i missread the grok pattern, i think its ok :)

PR #397, opened 2022-03-03
Check for 200 response code in http-bf-wordpress_bf filter

LePresidente:
Added check for 200 response code, as normal logins can be banned with the current rules


buixor (approved):

buixor:
great, thanks

PR #398, opened 2022-03-03
traefik parser - add traefik_router_name to statics

chad-jones:
Add traefik_router_name to statics so it is possible to see which service was targeted in alerts


woopstar:
+1

AlteredCoder (approved):

AlteredCoder:
Thanks for the PR! :)

PR #406, opened 2022-03-05
WIP: Add HTTP CVEs

woopstar:
Add multiple HTTP CVE's to the hub

### DIFF
--- a/collections/crowdsecurity/http-cve.md
+++ b/collections/crowdsecurity/http-cve.md
@@ -1,4 +1,4 @@
-A collection of http specific CVEs :
+A collection of http specific CVEs:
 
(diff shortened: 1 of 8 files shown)


he2ss:
Hi @woopstar,

Except `CVE-2017-9841`, we think that other scenarios you wrote can trigger some false positive because the http path can be used in a legit request. So be careful with those scenarios.
So If you want to keep those CVE scenarios, just remove them from the http-cve collection and keep them in new collection.

woopstar:
> Hi @woopstar,
> 
> Except `CVE-2017-9841`, we think that other scenarios you wrote can trigger some false positive because the http path can be used in a legit request. So be careful with those scenarios. So If you want to keep those CVE scenarios, just remove them from the http-cve collection and keep them in new collection.

Yes, we're considering to change it to a leaky bucket.

PR #409, opened 2022-03-06
Add Generic 401 Authorization Scenario to http-generic-bf

LePresidente:
Updated existing http-generic-bf scenario, with a  Generic 401 Scenario this will stop bruteforcing from a reverse proxy to Gotify for example using the 401 Http error code.

Updated test as well.


LePresidente:
If this should rather be a separate scenario let me know.

buixor:
Sorry for the delay, it looks rather good to me and should be merge very soon :)

buixor (approved):
lgtm

PR #410, opened 2022-03-06
Added seperate Generic 401 Http Error Scenario, this is an alternative to PR #409

LePresidente:
Added Generic 401 Scenario this will stop bruteforcing from a reverse proxy to Gotify for example using the 401 Http error code.


buixor:
Hey @LePresidente,

PR looks good to me, we're debating whether it should go in a more generic collection (ie. base-http-scenarios)

LePresidente:
I was debating with myself about the same thing, I did it first in nginx-http-generic-bf 
<https://github.com/crowdsecurity/hub/pull/409>

Though not sure if it would cause false positives and it will  become enabled by default since its already part of a generic collection

LePresidente:
Reopened the other PR as well since its possibly on the cards

buixor:
let's favor the other one, closing this one, thanks !

PR #411, opened 2022-03-07
Remove debug from authelia-bf, Fix filters for ombi,emby,authelia parsers so they only read there relevant logs.

LePresidente:
Remove debug from authelia-bf.yaml
Add the filter to the parsers so it only reads the files it should not all of them.


LePresidente:
Sorry about this, no idea why I removed this filter line but it kinda had a knock-on effect with all my contributions so far. This should fix it though

buixor (approved):
lgtm

AlteredCoder:
Hello, thanks for the PR. Can you just update the `type` of the example of acquisition in the markdown description file of the `emby` collection (cf. https://github.com/crowdsecurity/hub/blob/master/collections/LePresidente/emby.md) please ?

AlteredCoder (approved):

PR #423, opened 2022-03-30
Authelia-logs json parser could catch successful login attempts as bad.

LePresidente:
There is a possibility the JSON parser will return successful login attempts as bad attempts and ban the offending IP. This makes sure we check the entire string.


buixor:
Thanks, do you mind adding some test for the false positive so we can identify future regressions ?

buixor (changes requested):
lgtm, tests would be nice

LePresidente:
@buixor Ok I'm going to need some here here, The Grok Pattern looks to be correct but the Successful lines are still being detected by the Grok Pattern where it shouldn't be the case

LePresidente:
This is only happening for the Json Parsed Lines, The CLF log lines are working correctly.

LePresidente:
nevermind figured out the issue, I was not correctly understanding the json parser.

buixor (approved):
lgtm

PR #443, opened 2022-04-15
Update nginx-proxy-manager parser

baudneo:
The original was not parsing my logs as the upstream_status would be returned as '-'. The GROK for that was set to NUMBER, I changed it to DATA and it now parses my NPM logs again.

Example of my log which was not parsing before the change, upstream_status position is set as a '-' character.
```
[15/Apr/2022:01:33:52 -0600] - - 403 - GET https x.y.com "/vZ0zgF2G.wwwacl" [Client 1.2.3.4] [Length 171] [Gzip 3.23] [Sent-to 4.3.2.1] "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36" "-"
```


buixor (approved):
lgtm thanks

PR #444, opened 2022-04-16
Add GitLab parser, scenario and collection

timokoessler:
Adds GitLab parser, scenario and collection.

If anyone else finds a way to detect other failed authentications (for example: git pull), I'll be happy to add it.


parsers/s01-parse/timokoessler/gitlab-logs.yaml
```diff
@@ -0,0 +1,30 @@
+onsuccess: next_stage
+filter: "evt.Parsed.program == 'gitlab'"
+name: timokoessler/gitlab-logs
+description: "Parse GitLab Logs"
+pattern_syntax:
+  GITLAB_FAILED_PASSWORD: ^{"method":"POST","path":("\/users\/sign_in"|"\/users\/auth\/ldapmain\/callback").*("status":0|"action":"failure").*"time":"%{TIMESTAMP_ISO8601:timestamp}".*{"login":"%{USERNAME:username}".*"remote_ip":"%{IP:source_ip}".*
```
blotus, line 6: Parsing JSON with regexp seems a little fragile.
We have a `JsonExtract` helper that will parse JSON, it would be better to use it.
The doc is [here](https://docs.crowdsec.net/docs/next/expr/helpers/#jsonextractjsonblob-fieldname-string) and a usage example [here](https://github.com/crowdsecurity/hub/blob/1945256760861906cf1515e0192f60c444dd83cf/parsers/s01-parse/crowdsecurity/fastly-logs.yaml) (you can also just search for `JsonExtract` in this repository to see all the usage)

timokoessler: Okay, I will do that.
How can I combine multiple JsonExtract as a condition to set log_type?

parsers/s01-parse/timokoessler/gitlab-logs.yaml
```diff
@@ -1,30 +1,23 @@
...
     - meta: service
       value: gitlab
     - meta: source_ip
       expression: "evt.Parsed.source_ip"
     - target: evt.StrTime
       expression: "evt.Parsed.timestamp"
+    - meta: log_type
+      expression: "(JsonExtract(evt.Parsed.message, 'method') == 'POST' && (JsonExtract(evt.Parsed.message, 'path') in ['/users/sign_in', '/users/auth/ldapmain/callback'])) ? ((JsonExtract(evt.Parsed.message, 'status') == '0' || JsonExtract(evt.Parsed.message, 'action') == 'failure') ? 'gitlab_failed_password': (JsonExtract(evt.Parsed.message, 'status') == '200' && (JsonExtract(evt.Parsed.message, 'params') contains 'otp_attempt')) ? 'gitlab_failed_totp' : '') : ''"
+    - meta: username
+      expression: "(evt.Meta.log_type == 'gitlab_failed_password') ? (JsonExtract(evt.Parsed.message, 'params[1].value.login')) : ''"
```
timokoessler, line 23: Is there a better way than using params[1]?
Any suggestion on how to make the two lines more readable?

blotus:
Hello,

I tried to make the parser more readable, but currently crowdsec only has `JsonExtract` which returns a string, and makes it impossible to use some expr builtin functions (like `filter`) on the returned value.

I did a very quick & dirty POC with a new `JsonExtractSlice` function, which would give us something like this:
```
nodes:
  - filter: |
     JsonExtract(evt.Parsed.message, "method") == 'POST' && 
     JsonExtract(evt.Parsed.message, "path") == '/users/sign_in' && 
     any(JsonExtractSlice(evt.Parsed.message, "params"), {.key == 'user' && .value.login != ''})
    statics:
      - parsed: source_ip
        expression: JsonExtract(evt.Parsed.message, "remote_ip")
      - meta: log_type
        value: "gitlab_failed_password"
      - meta: username
        expression: filter(JsonExtractSlice(evt.Parsed.message, "params"), {.key == 'user' && .value.login != ''})[0]['value']['login']
```

I propose to put this PR on standby while we have a look at which helpers we should add to help parse more complex JSON objects, and I'll let you know as soon as we've added more helpers.

timokoessler:
Okay, that sounds good, thanks.

blotus:
Hello @timokoessler,

We've just merged new JSON helper to help parse complex JSON objects.

They will be available in 1.4.0, we can start working on this PR when it is released.

timokoessler:
Since I have a little less time in the next few weeks, I have already implemented the new JSON helper. I tested it with v1.4.0-rc1-46f2019. Of course, we still have to wait for the final release.

buixor (approved):
lgtm

blotus (approved):

blotus:
Thank you very much @timokoessler !

PR #446, opened 2022-04-17
Add sshesame collection

thespad:
A collection for the [sshesame](https://github.com/jaksi/sshesame) ssh honeypot.

### DIFF
--- a/collections/thespad/sshesame.yaml
+++ b/collections/thespad/sshesame.yaml
@@ -1,8 +1,6 @@
 parsers:
   - thespad/sshesame-logs
   - crowdsecurity/dateparse-enrich
-collections:
-  - thespad/sshesame
 scenarios:
   - thespad/sshesame-honeypot
 description: "Collection for sshesame SSH honeypot"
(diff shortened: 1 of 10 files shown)


he2ss (commented):
Tests seems broken. 
If I look to your sample logs and scenarios, I can say that it will generate more overflows in the scenario.assert. Maybe the previous one isn't updated ?

thespad:
Not sure how they got out of sync but I've updated them now and a local run is successful
```yaml
bash-5.1# cscli hubtest run sshesame
INFO[20-04-2022 01:47:02 PM] Running test 'sshesame'                      
INFO[20-04-2022 01:47:05 PM] Test 'sshesame' passed successfully (1724 assertions) 
------------------
 TEST      RESULT 
------------------
 sshesame  ✅     
------------------
```

collections/thespad/sshesame.yaml
```diff
@@ -0,0 +1,12 @@
+parsers:
+  - thespad/sshesame-logs
+  - crowdsecurity/dateparse-enrich
+collections:
+  - thespad/sshesame
```
he2ss, line 5: You can remove the collection itself as a dependency.

PR #450, opened 2022-04-19
Add CloudFlare V6 addresses to CDN whitelist

tomtom5152:
Fetched the same way as the V4 addresses. Method copied from `seo-bots-whitelist`.

Attempted to create tests for this but it seems post overflows aren't applied in `cscli hubtest` as no matter what I did I could not get it to apply even the unmodified version.


he2ss (approved):

he2ss:
Thanks for the contribution @tomtom5152 :+1:

tomtom5152:
Thought this would be a good one to get me started whilst exploring whitelists. I have a grand idea for a platform agnostic captcha bouncer in my head, but need to find the time for it yet.

PR #472, opened 2022-05-10
add f5 big ip cve 2022-1388

he2ss:


he2ss:
Maybe it's not a good Idea, because it can trigger some false positive.

PR #475, opened 2022-05-19
Add IPv6 to range postoverflow

AlteredCoder:
/!\ This need to be merge when crowdsec 1.4.0 is released.


buixor:
needs tests before being merged, but should be ok now that 1.4 has been  released

buixor (approved):
great

PR #478, opened 2022-05-25
add unifi support

blotus:
Add support for unifi logs:
 - New dropbear parser that plugs into the SSH BF scenario
 - New parser for unifi syslog log (we cannot use the standard syslog parser because the log is technically non-RFC compliant)
 - New unifi collection with the unifi parser, dropbear parser and iptables collection
 
 In order to work properly, this require https://github.com/crowdsecurity/crowdsec/pull/1554 to be released.


.tests/dropbear-ssh-bf/config copy.yaml
```diff
@@ -0,0 +1,10 @@
+parsers:
```
buixor, line 1: this one might be unwanted

buixor (changes requested):
minor change, otherwise LGTM

buixor (approved):

PR #492, opened 2022-06-20
Added Wireguard parser and scenario

klausagnoletti:
I've created my first PR ever! It's a logparsers for Wireguard and a scenario which detect two specific autentication errors.
I've followed the instructions in the doc so all testfiles are insluded. Looking forward to hear what you think :-)


buixor:
Hello @klausagnoletti 

Thanks for the effort. A few questions tho :

 - `Invalid handshake` : what is the risk that a random scanner/packet triggers this error (or the other one) while probing the service ?
 - `Packet has unallowed src IP` : what scenario would this match ? Is it truly some allowed list, or is it more a side-effect of failed auth ?

klausagnoletti:
Hey! Thanks for the comments. Those are excellent comments. 

As far as the `invalid handshake` this error can also happen in a legit setting so this should definitely be changed to a leaky scenario to avoid false positives and general annoyance. 

In `terms of packet has unallowed src ip` it's easier as indeed there is an allowed list define in your wireguard server settings which unsurprisingly contains ip of peers. This is from the `wg` manpage:

`AllowedIPs — a comma-separated list of IP (v4 or v6) addresses with CIDR masks from which incoming traffic for this peer is allowed and to which outgoing traffic for this peer is directed.`

I am not sure how well this list goes with roadwarriors on a dynamic ip. That should be investigated. I am talking to someone on the Wireguard reddit who might want to experiment a bit with it. In the meantime here's some more explanation of allowed ip.

https://utcc.utoronto.ca/~cks/space/blog/linux/WireGuardAllowedIPs

Is it possible to define two separate buckets in the same scenario file?

a1ad:
Hi @klausagnoletti what is the status? Do you use it in production? It would be nice to have some wireguard protection in opnsense.

PR #500, opened 2022-06-30
Opnsense openvpn added

testklaus:
This is a specialized parser and scenario for OpenVPN running on OPNsense. This is a work in progress as the most elegant solution would be to create a generic parser and scenario for OpenVPN regardless of platform. As described in the parser readme this depends on a few modifications to the syslog parser.


testklaus:
I did the PR with the wrong github account. D'oh!.

PR #505, opened 2022-07-08
parser, scenario and collection for suricata

buixor:


sabban:
Seems fine to me, but I would reflect that the scenario relies on the emerging threats ruleset in the name. The other way around we may tricks user to install useless stuff for them.

sabban (approved):
lgtm

PR #525, opened 2022-08-03
sshd-logs: match "Connection closed by invalid user"

maximelouet:
For sshd configurations that have an allowlist of users, sshd might log `Connection closed by invalid user` instead of `Connection closed by authenticating user` (even for users that do exist on the system).

Log-in attempts with these users are probably still malicious attempts, so we catch and block them.

Re-use the existing `SSHD_PREAUTH_AUTHENTICATING_USER` rule for this, to be consistent with `SSHD_PREAUTH_AUTHENTICATING_USER_ALT`.


LaurenceJJones (approved):
LGTM, passed using my own logs

Issue #389, opened 2022-02-18
Improvement to collection http-open-proxy


klausagnoletti:
This is a suggestion from the Discord community by Katze

Currently the proxy scenario only checks for the CONNECT method. I want to suggest that the following 3 possible methods could also be checked for:
```
37.59.135.136 - - [17/Feb/2022:12:55:01 +0000] "GET http://www.proxy-anonyme.com/gate.php HTTP/1.1" 400 657 "http://yandex.ru/" "Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US) AppleWebKit/534.16 (KHTML, like Gecko) Chrome/10.0.648.205 Safari/534.16" "-"
37.59.135.136 - - [17/Feb/2022:12:55:01 +0000] "\x04\x01\x00P%;\x87\x88\x00" 400 157 "-" "-" "-"
37.59.135.136 - - [17/Feb/2022:12:55:01 +0000] "\x05\x02\x00\x01" 400 157 "-" "-" "-"
```
the first is a different http proxy and the other 2 are the start of a socks4 and socks5 proxy connection.

some more technical information:
CONNECT is mostly used for https and the GET <url> are for http proxies, i don't know if that could add false positives.
SOCKS4 start with 0x04 and the second byte would be 0x01 or 0x02 (connect or bind) the rest determens the byte and ip so not useful for checking
SOCKS5 starts with 0x05 and after that it is the number of authentication methods supported, in the example 2 (probably user/pass and no auth) so its a bit harder but when a http connection starts with raw bytes it is already malformed.

katzeprior:
Will try to tackle this issue myself this summer vacation, need to make a new parser which I failed at last time I tried.

LaurenceJJones:
Linked to #855

PR #527, opened 2022-08-08
enhance http-open-proxy with get proxy

katzeprior:
As stated in this part of [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3.2) if the request is in the so called *absolute-form* it is a proxy.
This is mostly used for plain http connections, for https the *CONNECT* verb must be used.

With this pr I am trying to catch those proxy requests as well as stated in issue #389.
For the other 2 remaining proxy types (socks4/5) those require a whole new parser.

Hope I am not causing any issues and false positives with this.

Good Night

- Katze


LaurenceJJones:
LGTM :)

LaurenceJJones (approved):

PR #528, opened 2022-08-10
Add requested url to home assistant logs parser

libreshare:
On home assistant version 2022.8.2 there is a Requested URL area before the http user agent


LePresidente:
I think this should be a separate grok filter so both log formats are supported.

he2ss:
Hi, thanks for your contribution.

You need to remove the submodule hubtest and add the test to trigger the second overflow (with the url).

he2ss:
Hi, I updated the tests so we can merge this PR.

he2ss (approved):

PR #529, opened 2022-08-15
Update uptime-kuma-logs.yaml

joshaprior:
username can be whatever you want, which includes email addresses that break the pattern, changed USERNAME to DATA so it captures login attempts with emails as well.


buixor (approved):
lgtm

PR #543, opened 2022-09-06
Make Packet Filter Parser and Scan Scenario OPNsense compatible

Byter09:
Hello there!

I wanted to use this collection/parser/scenario and noticed that it doesn't quite work, so I figured I fix this :)

The only thing missing were some optional parameters, and some others had to be made optional. I took the opportunity to clean up the whole pattern design.

I do have a question about the last commit though. I can see, that this `service` meta info is being set in the `iptables` equivalent of this scenario, and was wondering why in this one? This should work perfectly fine for UDP. It's why I changed the service name. But if that has some importance, we can simply drop that commit.

Lastly, I have a question regarding the scenario.
Assuming my firewall gets port-scanned and thus this scenario blocks said IP, given how `CrowdSec` blocks on OPNsense, won't it still appear and continue to trigger the scenario? I think there needs to be some marker to differentiate between IPs that are fresh and those that have already been banned. Or I don't quite understand how this all works yet :smile: 

Anyway, I love OPNsense and CrowdSec, so here's my little contribution :)

EDIT: Nearly forgot to mention: I included syslog asserts because these two different log styles need to keep working in the future. :)


buixor:
Hello !

Thanks for the PR, we're having a look at it, and we'll come back to you soon :)

mmetc:
Hi @Byter09 !

I'm trying your changes, how should I configure the IDS/IPS service and how did you trigger the block (nmap options..)? Physical or virtual machine?

I was not able to trigger the blocks but I think it depends on my test environment.

Thanks

Byter09:
I don't quite follow the question. What I did was to have a look at the log output from the packet filter on my OPNsense firewall at
`/var/log/filter/latest.log`.

Did I forget to set that in this MR? I figured that the whole `filterlog` logic already reads this somehow. Sorry am not that versed in this tool yet :D

Zapfmeister:
If i understood correctly, this will make the pf firewall parser and bouncer compatible with opnsense? This would be awesome!
I would love to help on this. What would ne the best way to test this pr?

Byter09:
Hi, yes. That was the goal. I just changed the regex to make it compatible, however I'm unsure on how to proceed myself. I don't have that much time so I'd be happy if someone else could put the final touches on this or let me know exactly on how to get this merged.

LaurenceJJones:
Does these alterations help with #677

Byter09:
They should. I spend an evening testing my logs back then. So assuming nothing has changed much since then, it should just work. The main problem is that I don't really know how to get this merged. :(

a1ad:
@Byter09 is this parser still compatible with pfSense? Because that is the goal with the one I posted. in #677

Byter09:
I mean I left the original pfSense logs in there, so the goal is to have it be compatible, yes.

a1ad:
@LaurenceJJones what are we going to do? Do I need to test this PR or are we going forward with the one in 677?

LaurenceJJones:
> @LaurenceJJones what are we going to do? Do I need to test this PR or are we going forward with the one in 677?

Test this firstly just to ensure this is working and if so we can move ahead with this

a1ad:
@Byter09 nice work, it looks good. Just tested the TCP and UDP port scan successful. Going to test it a couple of days and with some more scenarios.
 
> I do have a question about the last commit though. I can see, that this `service` meta info is being set in the `iptables` equivalent of this scenario, and was wondering why in this one? This should work perfectly fine for UDP. It's why I changed the service name. But if that has some importance, we can simply drop that commit.

Can you answer this one @LaurenceJJones ? I can confirm UDP and TCP are working. Is it a problem the service name is tcp_udp instead of tcp? The scanarios are changed to tcp_udp as well.

```
statics:
  - meta: service
-  value: tcp
+ value: tcp_udp
```

> Lastly, I have a question regarding the scenario. Assuming my firewall gets port-scanned and thus this scenario blocks said IP, given how `CrowdSec` blocks on OPNsense, won't it still appear and continue to trigger the scenario? I think there needs to be some marker to differentiate between IPs that are fresh and those that have already been banned. Or I don't quite understand how this all works yet smile

When crowdsec bans an IP that IP is not logged anymore. So not in the logs and can't be parsed again.

Zapfmeister:
Can confirm working on opnsense! Nice job!

Byter09:
I'm glad nothing has changed since September to still have it work!

> When crowdsec bans an IP that IP is not logged anymore. So not in the logs and can't be parsed again.
But it would still appear in the filterlog as "blocked" right? Or does Crowdsec just ignore those filters then?

LaurenceJJones:
> Can you answer this one @LaurenceJJones ? I can confirm UDP and TCP are working. Is it a problem the service name is tcp_udp instead of tcp? The scanarios are changed to tcp_udp as well.

Should be no issue, you can name any key to any value. The reasoning behind the service flag is to use it within your scenario filters.

.tests/pf-logs/pf-logs.log
```diff
@@ -1,3 +1,8 @@
 Sep 28 10:38:09 pfSense filterlog: 53,,,1000000301,em0,match,block,in,4,0x0,,64,420,0,none,6,tcp,460,10.0.2.2,10.0.2.15,53742,22,420,PA,29633380:29633800,3617062173,65535,,
 Sep 28 10:38:24 pfSense filterlog: 53,,,1000102631,em0,match,block,in,4,0x0,,64,600,0,none,6,tcp,44,10.0.2.2,10.0.2.15,53792,22,0,S,31552001,,65535,,mss
-Sep 28 10:38:41 pfSense filterlog: 6,,,1000102434,em0,match,block,out,4,0x0,,64,0,0,DF,6,tcp,40,10.0.2.15,10.0.2.2,22,53742,0,FA,3617062173,29633380,65535,,
\ No newline at end of file
+Sep 28 10:38:41 pfSense filterlog: 6,,,1000102434,em0,match,block,out,4,0x0,,64,0,0,DF,6,tcp,40,10.0.2.15,10.0.2.2,22,53742,0,FA,3617062173,29633380,65535,,
```
LaurenceJJones, line 3: Are these log examples from opnsense? if so could we change the machine name from hostname to opnsense? so when checking the assert we can easily see which is opnsense and pfsense.?

Byter09: The first three lines are the original lines in that log. All I did was add a few of my OPNsense lines. This way we have a log containing a mix of pfSense and OPNsense logs. I was hoping that that would help upkeep it being compatible with both. It's just marked as changed because there was no new line before.

LaurenceJJones: Sorry meant the lines below that have `hostname` as the machine name. Would be nice if we could change them to `opnsense` then we know looking at the tests which is pfsense or opnsense

Byter09: I haven't touched this since September so lemme see if I can recall all the commands and stuff haha.

LaurenceJJones: I can do it if you like saves time

Byter09: I would *really* appreciate that. I honestly just made this to show a combined regex and was hoping someone could fully integrate it into crowdsec. :pleading_face:

PR #553, opened 2022-10-03
fix endlessh-logs parser to handle IPV4/IPV6 correctly

### LINKED ISSUE #445: endlessh: parser fails in ipv4-only setups
If endlessh is configured to only bind to IPv4 then the log format is
```
2022-04-16T15:06:56.918Z ACCEPT host=178.128.199.192 port=42832 fd=4 n=1/4096
```
Instead of
```
2022-04-16T15:14:17.341Z ACCEPT host=::ffff:129.151.93.41 port=42006 fd=4 n=1/4096
```
And therefore doesn't match
```
ENDLESSH_ACCEPT_V4: "%{TIMESTAMP_ISO8601:timestamp} ACCEPT host=::ffff:%{IPV4:source_ip} "
```

naimo84:
should fix #445

### DIFF
--- a/parsers/s01-parse/crowdsecurity/endlessh-logs.yaml
+++ b/parsers/s01-parse/crowdsecurity/endlessh-logs.yaml
@@ -3,8 +3,8 @@ filter: "evt.Line.Labels.type == 'endlessh'"
 name: crowdsecurity/endlessh-logs
 description: "Parse Endlessh logs"
 pattern_syntax:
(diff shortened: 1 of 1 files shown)


naimo84:
oh, I'm sorry that I didn't get it correctly... I've indeed set it to ipv4 only, so it's working for me now 🥳😉

many thanks for double checking 👍

LaurenceJJones:
I will look into seeing we can have best of both worlds. 👍🏻

LaurenceJJones:
Please give my changes a test but here grok debugger output
image

PR #556, opened 2022-10-05
Fix typo in all probing scenario occurences

mazzma12:


buixor:
postponed till the new hub version when we can rename items safely.

PR #563, opened 2022-10-10
Add Sonarr parser, scenario and collection

schiz0phr3ne:
Adds [Sonarr](https://github.com/Sonarr/Sonarr) parser, scenario and collection (installed with [this script](https://wiki.servarr.com/install-script)).


AlteredCoder:
Hello,

Thanks for your contribution! The scenario.assert file for "sonarr-bf" tests and the parser.assert file for "'sonarr-logs" are missing.

PR #593, opened 2022-11-13
Bind9 refused

mstilkerich:
This is a port of the fail2ban named-refused jail.

This is my first custom configuration for crowdsec, I appreciate feedback on things that can be improved!


collections/mstilkerich/bind9.md
```diff
@@ -0,0 +1,25 @@
...
+Acquisition configuration depends on whether bind9 is configured to log to
+syslog, separate log files, or both.
+
+For a separate log file, set the log type to `named`:
+
+```yaml
+filenames:
+  - /var/log/named/security.log
+labels:
+  type: named
```
LaurenceJJones, line 22: If type is set to syslog is the program name set the `named` within syslog?

parsers/s01-parse/mstilkerich/bind9-logs.yaml
```diff
@@ -0,0 +1,39 @@
...
+      pattern: '^%{BIND9_PREFIX}bad zone transfer request: ''\S+/IN'': non-authoritative zone (?:denied|\(NOTAUTH\))\s*$'
+      apply_on: message
+  - grok:
+      pattern: '^%{BIND9_PREFIX}(?:view (?:internal|external): )?query(?: \(cache\))? .* (?:denied|\(NOTAUTH\))\s*$'
+      apply_on: message
+statics:
+  - meta: service
+    value: bind9
+  - meta: log_type
+    value: bind9_denied
```
LaurenceJJones, line 31: This is okay at the moment, however, if we support multiple patterns they might need to be applied under each GROK pattern

LaurenceJJones (commented):
Pending above questions

collections/mstilkerich/bind9.md
mstilkerich: Yes, the bind daemon binary is called `named` and uses this also as the tag in syslog messages. Therefore, when log messages are taken from syslog the `syslog-logs` parser will set `Parsed.program` to `named`. Therefore I also use it as label for the bind9 logs, which the `non-syslog` parser will use for setting `Parsed.program` and subsequently the `bind9-logs` parser will see compatible events independent of whether they originate from syslog or a bind9 log file.

parsers/s01-parse/mstilkerich/bind9-logs.yaml
mstilkerich: I don't understand what you mean by supporting multiple patterns (there already are multiple patterns in the parser), but I also take you don't currently see an issue here. Anyway I saw no reason to tag the different policy violation event types differently and therefore all are tagged as `bind9_denied` type.

mstilkerich:
Thanks for your feedback, I have commented on your remarks.

Issue #575, opened 2022-10-24
Request : Add application-specific whitelists


buixor:
I'd like to see whitelists for know application's false-positives : 
 - [ ] nextcloud
 - [ ] matrix ?

ethindp:
Any update on this?

LaurenceJJones:
> Any update on this?

Can you be more specific? as you can see nextcloud is mostly flushed out but nobody has provided us with matrix logs / issues for us to be able to start one

ethindp:
Do you just need logs of non-200 status codes?

ethindp:
Or would you prefer just normal logs? Idk how stable the log output is but I can supply some.

LaurenceJJones:
Well just outline where the false positive is happening EG: "I get banned because too many 404 to /this/endpoints and it triggers these scenarios"

Then provide the logs that caused said statement above

PR #609, opened 2022-12-14
Nextcloud whitelist

LaurenceJJones:
When using a reverse proxy infront of nextcloud the contacts app has verbose 404 if "photo" is not found on .vcf file. This whitelist will prevent false positives with `crawl-non-statics`

Upstream https://github.com/nextcloud/contacts/issues/3021


LaurenceJJones:
Link to issue  #575

PR #610, opened 2022-12-16
Update radarr tests and add text file support

LaurenceJJones:


LaurenceJJones:
User confirmation the parser is working as intended.

he2ss (approved):

PR #612, opened 2022-12-18
Update web_php log parser

baudneo:
ZoneMinder has added user configurable PHP date formats. This covers legacy and US/CAN format, users with custom formats will need to modify the GROK pattern.


LaurenceJJones:
As always thank you @baudneo ! I will review it today.

LaurenceJJones:
Do think there be value adding
```
collections:
  - baudneo/zoneminder_http-cve.yaml
```
To the original zoneminder collections so if anyone has it installed originally they can automatically get the new cve collection without having to find it manually?

baudneo:
Yes that sounds like a good idea!

LaurenceJJones (approved):
LGTM

PR #613, opened 2022-12-19
use github oidc+role instead of key/secret for aws

nitescuc:
Avoid using aws persistent credentials and leverage OIDC instead


LaurenceJJones (approved):
LGTM

PR #622, opened 2022-12-29
Add 403 HTTP error code on http-generic-bf.yaml

thomaschampagne:
Unfortunately many applications may return "403 Forbidden" http code on authorization failure instead of "401 Unauthorized" http code. This `yaml` code simply adds this scenario which should be considered.


LaurenceJJones:
Could you get this branch up to date with master, so all we make sure all tests are passing.

thomaschampagne:
Hi @LaurenceJJones. Merge of master is done

PR #625, opened 2023-01-02
Tabobine

juliencouturier:


LaurenceJJones:
Is this all for a custom app you own or is the application available for everyone to use?

If so I think I saw you talking on discord and you wouldn't need to upload to the hub and I saw you have the answer to use a persistent volume

LaurenceJJones:
Thank you for spending the time on creating this, however, I don't believe this will suit the hub as it bespoke to your application. I will close for now if you feel I made a mistake you can reopen and question my opinion.

juliencouturier:
> Is this all for a custom app you own or is the application available for everyone to use?
> 
> If so I think I saw you talking on discord and you wouldn't need to upload to the hub and I saw you have the answer to use a persistent volume

Yes I do not need to upload those scenarios to the hub. 
Thank You

PR #628, opened 2023-01-08
add GET to 401 generic error

tasiotas:
Include GET requests that cause 401 errors

### DIFF
--- a/scenarios/crowdsecurity/http-generic-bf.yaml
+++ b/scenarios/crowdsecurity/http-generic-bf.yaml
@@ -18,7 +18,7 @@ type: leaky
 #debug: true
 name: LePresidente/http-generic-401-bf
 description: "Detect generic 401 Authorization error brute force"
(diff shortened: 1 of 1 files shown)


LaurenceJJones (approved):
lgtm

tasiotas:
@thomaschampagne I'm totally ok to roll it back.

I have Caddy server that has some protected routes, Caddy implements Basic auth via GET.
At first I made custom scenario to catch GET 401 which worked well, thought others might find it handy.

Btw. aren't your requests coming from private subnet (172.31.255.226) which usually is whitelisted?

thomaschampagne:
@tasiotas Rollback is a good option indeed. Maybe reconsider the scenario with different params with example I gave you. I can try to provide a scenario which match your needs if you want.

The public IP is `X.X.X.X` in my logs => I wanted to keep that private. She's not coming from a private subnet.

PR #630, opened 2023-01-08
Add collection for OpenVPN server installed on DD-WRT router

dawstudiocenter:


dawstudiocenter:
I've just updated some files to fix some failure during automatic tests.

dawstudiocenter:
Hi @LaurenceJJones,
I check and tests should be ok now.
Could you please take a look please at this PR ?
Thanks.

LaurenceJJones:
So the first issue we have is filtering on machine name? what if people dont have the same machine name?

We should fix the syslog parser to support dd-wrt syslog as the parser is "working" but it missing the program cause of the information in the middle

.tests/openvpnddwrt-logs/parser.assert
```diff
@@ -0,0 +1,187 @@
+len(results) == 3
+len(results["s00-raw"]["crowdsecurity/syslog-logs"]) == 12
+results["s00-raw"]["crowdsecurity/syslog-logs"][0].Success == true
+results["s00-raw"]["crowdsecurity/syslog-logs"][0].Evt.Parsed["logsource"] == "syslog"
+results["s00-raw"]["crowdsecurity/syslog-logs"][0].Evt.Parsed["message"] == "daemon.err openvpn[1422]: 5.181.98.144:32425 TLS Error: TLS handshake failed"
```
LaurenceJJones, line 5: Here the example of raw syslog not getting it correctly.

The message should only be `5.181.98.144:32425 TLS Error: TLS handshake failed`

dawstudiocenter: Sorry I do not understand.
the machine name (or IP) is not used to trigger an event. I only want to trig an event on  "TLS Error : TLS handshake failed" pattern or "TLS Error : TLS key negotiation failed to occur within 60 seconds" pattern in syslog message.
It's the reason why I use such pattern in parser yaml file : 
pattern_syntax:
  OPENVPNDDWRT_BADKEY: '^%{DATA:program}: %{IP:ip}:%{POSINT:port} TLS Error: TLS key %{DATA:error}$'
  OPENVPNDDWRT_BADHS: '^%{DATA:program}: %{IP:ip}:%{POSINT:port} TLS Error: TLS handshake failed%{DATA:error}$'

Here are some examples of syslog messages as recorded on DD-WRT software (build 49544) : 
Jan 22 06:50:36 VPNRouter daemon.err openvpn[1409]: 94.102.61.34:35806 TLS Error: TLS key negotiation failed to occur within 60 seconds (check your network connectivity)
Jan 22 06:50:36 VPNRouter daemon.err openvpn[1409]: 94.102.61.34:35806 TLS Error: TLS handshake failed

I did not change the configuration of openvpn log message format.

dawstudiocenter: May be, there's something that I did not understand in parser pattern usage. Let me know.

parsers/s01-parse/dawstudiocenter/openvpnddwrt-logs.yaml
```diff
@@ -0,0 +1,29 @@
+name: dawstudiocenter/openvpnddwrt-logs
+description: "Parse ddwrt logs for openvpn messages"
+filter: "evt.Meta.machine == 'VPNRouter'"
```
LaurenceJJones, line 3: Not everyone machine will be named `VPNRouter` ?

.tests/openvpnddwrt-logs/parser.assert
LaurenceJJones: `the machine name (or IP) is not used to trigger an event` it is your filtering at the top to make sure the machine name is `VPNRouter`? I triggered another comment.

LaurenceJJones: There is nothing wrong with your parser inparticular, it just the our syslog parser is not expecting the `daemon.err` as that not standard syslog logs

dawstudiocenter: VPNRouter is the name of my router. 
I do not mention it in my parser pattern. Or it is not my intention.
But you're right VPNRouter is present in the log file and assert file I use for the tests. May be I'm wrong but I understand that I need to provide some kind of syslog message in sample parser log file. it is real example.

parsers/s01-parse/dawstudiocenter/openvpnddwrt-logs.yaml
dawstudiocenter: OK I understand now. You're right. I filter syslog message using my router name.
I need to remove this filter condition or adapt it.
Let me think about it .

dawstudiocenter:
I just change the filter condition to match on syslog message content instead of host name (my router VPNRouter).
The tests assert files have been run and updated accordingly.

LaurenceJJones:
I will allocate some time early next week, to offer up a more optimal fix for this. Just wanted to make it clearer there is nothing you done that is wrong, if we make some changes to our syslog parser we could make it fit for a generic openvpn collection.

dawstudiocenter:
Hi @LaurenceJJones , any news on your side ? let me know If i can help.

PR #632, opened 2023-01-13
Allow Teamspeak Parser to deal with IPv6 IPs

Xyaren:


LaurenceJJones:
Just so the functional tests are better could we add this line to the .tests log and redo the parser assert?

This will ensure any changes moving forward will not affect this compatibility.

Xyaren:
Just pushed something, can you check please ?
I did not run it locally.

PR #653, opened 2023-02-13
add meshcentral collection to hub

a1ad:
Adds support for Meshcentral

Note:
In order to have auth messages, we need to set "authLog": "/opt/meshcentral/meshcentral-data/auth.log" in the Meshcentral config file.

This is my first PR here, so be sure to give it a thorough check. Used what I learned from @LePresidente jellyfin PR.


LaurenceJJones:
I will check this today! Thank you for spending the time creating this we do appreciate time spent. 🎆

LaurenceJJones:
Please upload the assert that goes to screen when running hubtest to .tests/meshcentral-bf/scenario.assert

`sudo cscli hubtest run meshcentral-bf`

a1ad:
oh i was under the impression that the hubtest did that automagically

LaurenceJJones:
> oh i was under the impression that the hubtest did that automagically

it does run them but it needs the test file to compare the output if not provided the check will always fail

PR #656, opened 2023-02-14
Make custom parser pattern syntax unique.

LePresidente:
Issue found on discord,

If the Pattern Syntax has the same name for multiple parsers, the first one is always used and not replaced. 

So for example the first CUSTOMDATE will be reused instead of being unique to each parser


LePresidente:
Sorry about that should be sorted.

PR #667, opened 2023-02-25
Added match for SSH Failed authentication attempt

jahanson:
Added a match pattern to also pick up SSH Failed authentication attempts and not just the web failed authentication attempts.

The new grok pattern will match:
2023/02/25 20:26:14 modules/ssh/ssh.go:281:sshConnectionFailed() [W] [63fa641c-19] Failed authentication attempt from 1.2.3.4:57094

### DIFF
--- a/parsers/s01-parse/LePresidente/gitea-logs.yaml
+++ b/parsers/s01-parse/LePresidente/gitea-logs.yaml
@@ -13,6 +13,12 @@ nodes:
       statics:
         - meta: log_type
           value: gitea_failed_auth
+  - grok:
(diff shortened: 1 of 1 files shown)


LaurenceJJones:
Just so I understand added SSH failed attempts to gitea parser? does gitea also log ssh attempts to git?

jahanson:
Yes, gitea implements ssh and installing it usually means putting the server's ssh on a different port. Here are some links to the docs to confirm:
https://docs.gitea.io/en-us/install-with-docker-rootless/
https://docs.gitea.io/en-us/logging-configuration/#debugging-problems

PR #668, opened 2023-02-27
Split generic bf to separate scenarios

LePresidente:
Moved 401 and 403 generic bf to its own scenario, 

alot of arr apps return 403 for everything when doing api calls so its causing alot of false positives. 

Moving this to its own scenario will allow people who don't want it to ban to move it to a simulation instead

cscli simulation enable crowdsecurity/http-generic-403-bf

Added tests for 403 error as well


LePresidente:
I guess another option would be to allow the simulation command to work on sub scenarios in the scenario file instead of just the main file.

LaurenceJJones:
Just to know I havent forgot about this PR. We discussed it internally and we believe it people dont delete the files they will get duplicate scenarios as calling `hub update && hub upgrade` will never delete any files.

PR #673, opened 2023-03-06
Freeswitch support

LaurenceJJones:


collections/crowdsecurity/freeswitch.md
```diff
@@ -0,0 +1,15 @@
+## freeswitch collection
+
+### Description
+
```
buixor, line 4: I think we need to say what the collection covers

buixor (changes requested):
minor suggestion but LGTM otherwise

buixor (approved):
LGTM

PR #678, opened 2023-03-13
Adds Redmine Collection to hub.

LePresidente:
Adds Redmine Collection to hub.


LaurenceJJones (approved):
As always great work @LePresidente 🙏🏻

PR #696, opened 2023-03-30
Parser and Scenario for mailu admin container

mwinters-stuff:
Parser and Scenario for mailu's admin container - for finding authentication attempts.

2nd attempt at PR - first was from the wrong branch which didnt have all the commits.


LaurenceJJones:
I think your tests might be out of date, just update the .assert files

mwinters-stuff:
assert files updated.. :)

PR #697, opened 2023-03-31
Fix Nextcloud false positives

GNU-Plus-Windows-User:
This pull request fixes 2 false positives with Nextcloud and CrowdSec
1. Images are not marked as images by Nextcloud (not ending in jpg,png etc) and can easily trigger HTTP crawl non statics scenario.
2. 404 error is thrown when opening a folder and ``readme.md`` doesn't exist, in rare cases this can trigger HTTP Probing if a user quickly opens multiple folders.


GNU-Plus-Windows-User:
I've found an additional false positives for files that are located in the trash bin, 404 errors are thrown when previews are missing and triggers http probing.

PR #716, opened 2023-04-27
Add AWS ALB access log parser

Zaulao:
Adds a parser for Application Load Balancers (ALB) access logs (which is often used as ingress in Kubernetes clusters).

From the raw log, relevant fields for HTTP analysis are extracted. I used the parser for Nginx and the data analyzed by some of the base-http-scenarios as a basis.


Zaulao:
No problem! Just added it

Zaulao:
Sorry for the typo, I got it wrong after forking the project to prepare de PR

.tests/aws-alb-logs/config.yaml
```diff
@@ -0,0 +1,11 @@
+parsers:
```
LaurenceJJones, line 1: can we add s00 stage parser

`- crowdsecurity/syslog-logs`

Even though it is not really used it "real runtim" there is always a stage s00 parser

.tests/aws-alb-logs/parser.assert
```diff
@@ -0,0 +1,286 @@
+len(results) == 3
```
LaurenceJJones, line 1: Once you add stage s00 parser you will need to regen .assert

LaurenceJJones (changes requested):
Small changes but we also need tests for some scenarios. If you can bring up to upstream changes and also do my requested changes I can create the scenarios tests.

Zaulao:
Is there anything else I can do here?

blotus (approved):

blotus:
Hello,

Sorry about the delay; the PR fell through the cracks.

I'm merging it; thanks for the contribution!

PR #730, opened 2023-05-16
Added harbor collection

LePresidente:
Added basic [Harbor](https://goharbor.io/) collection to Hub.


LaurenceJJones:
@LePresidente Does this have the latest master branch changes? as we fixed the bitwarden

LePresidente:
Hi, according to github my branch is only one commit ahead of master so yes it should be up2date

PR #731, opened 2023-05-16
Fix Nextcloud false positive when not using pretty URLs

GNU-Plus-Windows-User:
In the CrowdSec Discord self hosting channel, a false positive was reported with Nextcloud when loading image previews, the whitelist assumes the user is using pretty URLs. This PR fixes that false positive.


LaurenceJJones (commented):
Might be because I'm on mobile, but is there extra `==` next to the `contains`?

GNU-Plus-Windows-User:
Yes there is ``& evt.Parsed.request contains == '/core/preview``, I assume you want me to remove it?

PR #733, opened 2023-05-19
Change iptables to use new 1.5 kv parser and update scenario 

LaurenceJJones:
since we always hardcoded to tcp didnt allow users to have flexibility.


.tests/iptables-logs/parser.assert
```diff
@@ -1,56 +1,236 @@
...
+results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["datasource_type"] == "file"
 results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["log_type"] == "iptables_drop"
 results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["machine"] == "ubuntu-focal"
-results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["service"] == "tcp"
+results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["service"] == "udp"
 results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["source_ip"] == "127.0.0.1"
-results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["datasource_path"] == "iptables-logs.log"
-results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Meta["datasource_type"] == "file"
\ No newline at end of file
+results["s01-parse"]["crowdsecurity/iptables-logs"][3].Evt.Unmarshaled["iptables"]["OUT"] == "MAC"
```
LaurenceJJones, line 121: OUT should be empty??

scenarios/crowdsecurity/iptables-scan-multi_ports.yaml
```diff
@@ -1,7 +1,7 @@
 type: leaky
 name: crowdsecurity/iptables-scan-multi_ports
 description: "ban IPs that are scanning us"
-filter: "evt.Meta.log_type == 'iptables_drop' && evt.Meta.service == 'tcp'"
+filter: "evt.Meta.log_type == 'iptables_drop' && evt.Meta.service in ['tcp', 'udp']"
```
LaurenceJJones, line 4: Since we always hardcoded TCP we did detect UDP drops as well.  Currently I have just replicated the current behaviour but we may want to only look for TCP

buixor (approved):
lgtm

PR #737, opened 2023-05-20
Fix jellyfin logs

gauth-fr:
When using logs from Docker logs (directly or forwarding them to a syslog server), quotes around name and IP are missing.
This PR fixes this.

Also, when using docker logs, (so the container console logs), there is no date, just the time, and so logs won't match.
I edited the documentation file to provide guidance on how to fix this and have the date in the logs.


LaurenceJJones:
can you merge upstream changes into your branch

gauth-fr:
Ok i think I did it, hopefully correctly :)

PR #738, opened 2023-05-20
Add navidrome collection, scenario & parsers

gauth-fr:
Support for [navidrome](www.navidrome.org) logs


LaurenceJJones:
Same again please merge upstream changes.

gauth-fr:
Again, hope it did it correctly!

LaurenceJJones:
Seems your scenario assert may be empty?

gauthiersiri:
Hi again @LaurenceJJones, sorry for the delay, but i think this is the good one this time...?
Crossing fingers ^^

parsers/s01-parse/gauth-fr/navidrome-logs.yaml
```diff
@@ -0,0 +1,29 @@
...
+#debug: false
+name: gauth-fr/navidrome-logs
+description: "Parse navidrome logs"
+filter: "evt.Parsed.program == 'navidrome'"
+pattern_syntax:
+  NAVIDROME_CUSTOMUSER: "(%{EMAILADDRESS}|%{USERNAME})"
+  NAVIDROME_CUSTOMDATE: "%{YEAR}-%{MONTHNUM}-%{MONTHDAY}T%{HOUR}:%{MINUTE}:%{SECOND}Z"
+nodes:
+  - grok:
+      pattern: 'time="%{NAVIDROME_CUSTOMDATE:timestamp}".*msg="Unsuccessful login".*X-Forwarded-For:\[%{IP:source_ip}\].*username=%{NAVIDROME_CUSTOMUSER:username}'
```
LaurenceJJones, line 11: What if the user does not use X-Forwarded-For header?

gauth-fr: I won't lie i don't know :p
I guessed any reverse proxy would provide this header...
I tried thru direct IP, but in my case, these log line would not show any IP at all.
But another line would eventually show an IP. 
```time="2023-07-20T19:21:29Z" level=warning msg="HTTP: POST http://192.168.0.250:4533/auth/login" elapsedTime=1.2ms httpStatus=401 remoteAddr="172.19.0.1:52626" requestId=8c5a220df7c3/SsMOTbrPD6-018150 responseSize=40```
But in my case, since im running navidrome as a container, i just get the docker network gateway which is useless.

So any guidance is welcome actually... :)

Comment (LaurenceJJones) [reply parsers/s01-parse/gauth-fr/navidrome-logs.yaml:11]: > I guessed any reverse proxy would provide this header...
It is best practice to pass this through, not mandatory. 

hmmm I wouldn't be happy to merge if this wasnt an option inside navidrome to set an actual trusted reverse proxy. Looking at configuration it is not possible. I will convert this to a draft and when I have more time to spend on booting up a copy and check it myself then we can explore a merge

**Comment by LaurenceJJones** [comment]
Converting to draft as parser is expecting to receive a proxied header means two things:

- Application does not handle trusted proxies so can be spoofed if user exposes the port unknowingly
- If application is exposed via port directly parser will not work as it intends to find header

**Comment by gauth-fr** [comment]
No problem ;) If you need anything feel free to ask!

**Comment by sdwilsh** [comment]
Navidrome now has a `ReverseProxyWhitelist` setting, and we also don't have to rely on the headed in logs anymore.  For example, I now see this when I fail login: 
```
time="2025-01-30T05:07:37Z" level=warning msg="HTTP: POST http://navidrome.tswn.us/auth/login" elapsedTime="792.538µs" httpStatus=401 remoteAddr=10.117.0.31 requestId=navidrome-statefulset-0/YzKDTM4yVE-001467 responseSize=40 userAgent="Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:134.0) Gecko/20100101 Firefox/134.0"
```

I think that satisfies @LaurenceJJones's concerns but did require updates to this PR.

sdwilsh:
@LaurenceJJones, I ended up opening a new PR for this that I think addresses your feedback here in #1259.

PR #749, opened 2023-05-30
Generalizes Exim patterns

ppkarwasz:
This PR improves Exim patterns on 3 aspects:

 * Exim's `$sender_fullhost` is of the form `<hostname> (<helo>) [<ip>]:<port>`, where only `<ip>` is always present.  The presence of `<port>` depends on the `incoming_port` configuration option. The presence of `<hostname>` depends on a reverse IP lookup.  The `<helo>` is printed only if it differs from `<hostname>`.
 * Adds optional TLS data when applicable,
 * Works with Syslog.

The previous GROK patterns missed some of these possibilities. They also mistakenly marked the `source_dns` and `source_helo` as `target_dns` and `target_helo`.

More details about Exim logs can be found in the [Log Files chapter](https://www.exim.org/exim-html-current/doc/html/spec_html/ch-log_files.html). of the documentation.


ppkarwasz:
@buixor,

Ooops, I forgot about that. I corrected the `debug: true` and rebased the PR onto the current `master`.

LaurenceJJones (approved):
LGTM merged

PR #771, opened 2023-07-01
Paperless ngx

andreasbrett:
Add [Paperless-ngx](https://github.com/paperless-ngx/paperless-ngx) collection incl. parser and scenarios to the hub


.tests/paperless-ngx-logs/parser.assert
LaurenceJJones, line 1: Could you include all of the parser assert output not just for s01-parse stage

LaurenceJJones (changes requested):
Other than single comment once that is done, can be merged

Also thank you for spending the time making this! 🎆

.tests/paperless-ngx-logs/parser.assert
andreasbrett: fair point. latest commit includes all the parser assert output.

LaurenceJJones (approved):
LGTM flakey tests (outside scope of this PR)

PR #774, opened 2023-07-06
Add paloalto threat log

AlteredCoder:


collections/crowdsecurity/palo-alto.md
```diff
@@ -0,0 +1,17 @@
+## Palo Alto collection
+
+This collection support:
```
buixor, line 3: maybe add something like : it's expected for the admin to forward palo alto logs to an existing facility and read the logs of palo alto with crowdsec

buixor (changes requested):

buixor (approved):
lgtm

PR #778, opened 2023-07-18
Support Jellyfin Using Syslog

LePresidente:
Support Jellyfin Using Syslog


LaurenceJJones (approved):
LGTM wait for hubtests to run I know about flakkey tests

PR #782, opened 2023-07-27
Fix FP with Nextcloud desktop client

GNU-Plus-Windows-User:
Fixes false positive when syncing files from the nextcloud desktop client


LaurenceJJones:
Know its a long shot do you have any sample log lines? that you can add to `nextcloud-whitelist` folder.

In a new version of crowdsec we added the whitelist to the output so if we preemptive get them now we have some data.

GNU-Plus-Windows-User:
@LaurenceJJones I'll add it if I can reproduce the FP.

GNU-Plus-Windows-User:
@LaurenceJJones I was only able to reproduce one of them. I have the ``cscli alerts inspect -d`` output for the other is that fine?

LaurenceJJones:
> @LaurenceJJones I was only able to reproduce one of them. I have the `cscli alerts inspect -d` output for the other is that fine?

hmm might be able to reconstruct the stuff so best to show it and I can work on it 👍🏻

GNU-Plus-Windows-User:
@LaurenceJJones I've just encountered the false positive myself, I've added some logs to the test. I've fixed a new FP with Nextcloud 28 and newer as well.

PR #783, opened 2023-07-27
add support for webmin

andreasbrett:
add support for [webmin](https://github.com/webmin/webmin)


LaurenceJJones:
Just FYI github actions are UTC timestamp and marshaltime in enrich will unmarshal EPOCH to your timezone so you need to run `cscli hubtest` with these set

`sudo TZ=UTC cscli hubtest <your_test>`

andreasbrett:
how would that work on 

> Just FYI github actions are UTC timestamp and marshaltime in enrich will unmarshal EPOCH to your timezone so you need to run `cscli hubtest` with these set
> 
> `sudo TZ=UTC cscli hubtest <your_test>`

andreasbrett:
Sorry if I'm being pushy but could this please be merged?

PR #785, opened 2023-07-31
Remove the IP matching workaround

vamega:
The workaround is no longer necessary, as the issue was fixed by forking grokky (https://github.com/crowdsecurity/crowdsec/issues/936)

### DIFF
--- a/parsers/s01-parse/crowdsecurity/sshd-logs.yaml
+++ b/parsers/s01-parse/crowdsecurity/sshd-logs.yaml
@@ -4,18 +4,14 @@ filter: "evt.Parsed.program == 'sshd'"
 name: crowdsecurity/sshd-logs
 description: "Parse openSSH logs"
 pattern_syntax:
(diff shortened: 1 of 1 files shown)


vamega:
I have no idea how `aws-cloudtrail-postexploit` is failing in the CI run.
That works on my machine.

Looking at the `config.yaml` file for the test, it doesn't seem possible that this change could cause that test to fail.

LaurenceJJones:
> I have no idea how `aws-cloudtrail-postexploit` is failing in the CI run.
> That works on my machine.
> 
> Looking at the `config.yaml` file for the test, it doesn't seem possible that this change could cause that test to fail.

Don't worry about it. It's a race condition. we have a fix merged. We just need to fix CI

LaurenceJJones:
Closing this workaround is indeed not really needed but there are merged conflicts.

If you indeed thing this is an issue that causing problems then please reopen PR with the conflicts resolved.

PR #788, opened 2023-07-31
add Provider and Product to the sysmon parser

buixor:


LaurenceJJones (approved):
LGTM

parsers/s01-parse/crowdsecurity/sysmon-logs.yaml
```diff
@@ -5,6 +5,8 @@ description: "Parse sysmon events"
 nodes:
   - filter: evt.Parsed.EventID == '1'
     statics:
+      - parsed: ProviderName
+        expression: XMLGetNodeValue(evt.Line.Raw, "/Event/System/Provider/@Name")
```
blotus, line 9: It should be `XMLGetNodeValue(evt.Line.Raw, "/Event/System[1]/Provider/@Name")`

blotus (changes requested):

PR #796, opened 2023-08-04
lolbins detection from sigmahq

buixor:


buixor:
tested on upsttream 1.5.2 and it works (ping @AlteredCoder )

AlteredCoder (approved):

PR #800, opened 2023-08-11
Readd wireguard PR

LaurenceJJones:
@klausagnoletti did most of the work so thank you I jut did some minor tweaks:

- All tests include dateparse-enricher by default to ensure the timestamp is compliant
- Auth bruteforce is now leaky since we never got confirmation if scenario can trigger FP's so this sets some sort of default
- since wireguard is UDP based it hard to find the service since it does not respond by default.
- Dropped @klausagnoletti name from parser/scenario so if any users has issues they can come to us

one potential clash is iptables parser since iptables does not know a wireguard line from another kernal entry so will investigate


klausagnoletti:
Wow. That's the first time my coding skills have ever been mentioned in a positive tone. I'm happy that my works is of any use albeit it's quite difficult to use and questionable if anyone would ever set up Wireguard to log in a production environment. But whatevs 🤡

LaurenceJJones:
> it's quite difficult to use

Yeah I spent 30 minutes trying to get it to log anything. So I added a quick disclaimer at the top because its usable but might not be practical

PR #802, opened 2023-08-12
Add Collection: Cowrie Honeypot

Kornelius777:
This collection adds the ability to discover Code Execution Attempts within the Cowrie Honeypot.


LaurenceJJones:
Hey thank you for your PR

So the first issue I see is we already have a parser for cowrie connection logs https://github.com/crowdsecurity/hub/blob/4ddce084e6bc313033cbfb2753d38f4569a0ced5/parsers/s01-parse/crowdsecurity/cowrie-logs.yaml#L4

And it uses the same key I would suggest merging these into the same parser file and crowdsecurity one has existed longer so that would be the target

PR #818, opened 2023-09-02
fix(paperless-ngx): Remove ^ from the patterns to allow container log…

dimatha:
First of all - Thanks for the great work ! 

It seems like we have to remove the start of the line matching filter to allow container logs to be parsed.


dimatha:
@andreasbrett, could please have a look ? Thanks !

andreasbrett:
> @andreasbrett, could you please have a look ? Thanks !

LGTM. I think we could also get rid of the trailing `$`.

andreasbrett (approved):
still works and I think we should also get rid of the trailing `$`. there's no need to have this regex utterly strict.

dimatha:
@andreasbrett  done ! 

Thanks

LaurenceJJones (commented):
LGTM

PR #819, opened 2023-09-02
fix(authelia): Remove ^ from the patterns to allow container log parsing

dimatha:
First of all - Thanks for the great work !

It seems like we have to remove the start/end of the line matching filter to allow container logs to be parsed.

Similar to https://github.com/crowdsecurity/hub/pull/818


dimatha:
Hello @LePresidente ! Could you please have a look ? Thanks

LaurenceJJones:
Seems you have some failing tests? Seems the overflow didn't happen at all

dimatha:
@LaurenceJJones I should sleep during the nights :)
Now is it failing on 'CVE-2022-35914' ?

PR #825, opened 2023-09-06
Add sshd impossible travel

LaurenceJJones:


andreasbrett:
When can we expect to see this getting merged? I really dig advanced detections like these. Making use of enriched metadata that way could introduce some very interesting scenarios.

PR #826, opened 2023-09-08
Update http-probing.yaml

SDx3:
"ressource" > "resource"


LaurenceJJones:
This change will break the scenario as you would also need to change where this variable is defined

SDx3:
Good point, I didn't know. So where else is this misspelled, do you happen to know? Or are we at the point where "http_referer" is now as well?

LaurenceJJones:
It is defined here

https://github.com/crowdsecurity/hub/blob/8083cd0c7336cde63b5feb784f3e090c171976c9/parsers/s02-enrich/crowdsecurity/http-logs.yaml#L32-L33

PR #833, opened 2023-09-24
Adds Bookstack & Joplin server parser, scenarios and tests

xs539:
parser/scenario & test for [Bookstack](https://www.bookstackapp.com)
parser/scenario & test for [Joplin Severr](https://github.com/laurent22/joplin/tree/dev/packages/server)


LaurenceJJones:
LGTM, however, can we copy the parsers readme to collection level as well as users may not load the parser to see the example config.

xs539:
@LaurenceJJones  Done!

PR #834, opened 2023-09-30
Enhance CrowdSec Collections with GitHub IP Whitelist

Psycho0verload:
In environments with high GitHub interaction, particularly where GitHub Actions are heavily utilized, it's common to encounter issues with GitHub being inadvertently blocked. This leads to GitHub Actions failing unexpectedly. To mitigate this issue, I've compiled a whitelist containing all official GitHub IP addresses, sourced from GitHub's Meta-API.


LaurenceJJones:
Hey thanks for the PR, the only issue we can see this pretty much whitelist the whole of Azure ranges. So if you get attacked from a rouge VM on azure you will just whitelist it by default.

I would love to understand the issues and could it not be accomplished with using something a like VPN from actions to your infra?
https://medium.com/@mahmutali.mas/access-private-server-with-github-actions-vpn-connection-cf20313abdd5

Psycho0verload:
Thank you @LaurenceJJones for the prompt review and feedback on my PR. After submitting it, I did give some thought to the issue you pointed out regarding the potential for attacks originating from Azure networks, or even GitHub Actions themselves. I understand and appreciate your concern.

The article you shared offers a smart alternative solution to bypass this issue by using a VPN. It presents a well-thought-out way to ensure that only trusted internal ranges can access our infrastructure. I'd like to thank you for bringing this to my attention; it provides a different angle on solving the problem that I will definitely consider implementing.

PR #835, opened 2023-10-04
Add scenario to detect suid binaries crashing

blotus:


scenarios/crowdsecurity/auditd-suid-crash.yaml
```diff
@@ -0,0 +1,20 @@
+type: conditional
+name: crowdsecurity/auditd-suid-crash
+description: "Detect root suid process crashing"
+filter: |
+  (evt.Meta.log_type == 'execve' && evt.Meta.euid == '0' && evt.Meta.auid != '0') ||
```
LaurenceJJones, line 5: Wont this line detect all SUID execution no matter the status code?

blotus: That's intended, we cannot know in advance if a process will crash or not, and we do group by pid, meaning each process will get their own bucket

buixor (approved):

PR #838, opened 2023-10-05
Update gitea-logs.yaml

R-Nabil:
Hi All, 

Hope i'm doing this the right way, but I have noticed that the gitea log parser doesn't catter for the case of a user's password being invalid multiple time in a row (bf scenario use case). That PR is what I have test locally as a modification of the scenario to "fix" that.

Let me know if that is not the appropriate way to suggest a fix.

One other approach could be to make the first pattern more generic with something like :

pattern: "^%{GITEA_CUSTOMDATE:timestamp}.*?Failed authentication attempt for %{GITEA_CUSTOMUSER:username} from %{IP:remote_ip}:%{NUMBER:remote_port}.*?"

Thks,

Example Log : 
#line: 2023/10/05 16:48:46 ...ers/web/auth/auth.go:206:SignInPost() [I] Failed authentication attempt for nabil from a.b.c.d:0: user's password is invalid [uid: 1, name: xxxx]


LaurenceJJones:
Hey 👋🏻 

Thank you for taking the time to do this PR! dont suppose you have an example log line for this case? if so we can add it to the test and ensure compatibility moving forward. 

Please remove/redact any information you dont want to show example user `bob` is valid but I dont want to leak that so I will edit the line to be `toto`

Dont worry about generating the test I can do that for you, I just need the log line

R-Nabil:
Hey,

Not a problem. 

I am not an expert so I also suggested in my comment to either add the new pattern or amend the first one to be more generic. As I am not sure what's the favorite approach I duplicated it for now. 

Let me know if there's anything else I can do

LaurenceJJones:
> Hey,
> 
> Not a problem.
> 
> I am not an expert so I also suggested in my comment to either add the new pattern or amend the first one to be more generic. As I am not sure what's the favorite approach I duplicated it for now.
> 
> Let me know if there's anything else I can do

No worries, I just pushed tests and there was an ident issue can you confirm this works locally for you?

LaurenceJJones (approved):
LGTM, dont worry about broken tests on suid its nothing you done

PR #848, opened 2023-10-10
 L7 DoS tools detection

buixor:
Add a collection and associated scenarios to detect some well-known basic HTTP Dos tools seen recently. You know what I mean.


LaurenceJJones (approved):
LGTM

PR #868, opened 2023-11-06
add confluence cve-2023-22518

he2ss:


LaurenceJJones:
imo the confidence should be lower than 3 since we dont know if they is commonly used and is apart of confluence

blotus (approved):

PR #870, opened 2023-11-09
 bind9: also block 'query failed (REFUSED)' 

sylvestre:
+ a small ride along


LaurenceJJones:
Can you also include the generated tests?

PR #873, opened 2023-11-14
added cs-cloudflare-worker-bouncer to blocker list

rr404:


rr404:
already existing : https://github.com/crowdsecurity/hub/pull/872

PR #875, opened 2023-11-21
Add a test for MariaDB and fix typos + MariaDB collection docs

williamdes:
```log
2023-11-21 19:42:20 134334 [Warning] Access denied for user 'root'@'34.140.248.32' (using password: NO)
```

https://mariadb.com/kb/en/server-system-variables/#log_warnings

```ini
log_warnings=3
```


LaurenceJJones:
Could we have a test log line and update asserts?

williamdes:
> Could we have a test log line and update asserts?

In fact I noticed that I should have looked at MariaDB, I did a mistake.
Anyway, here is a new test for a non local IP

williamdes:
I fixed the documentation, everything works now

PR #887, opened 2023-12-19
Update nextcloud-whitelist.yaml for non-existent previews in the trashbin

ConzzTV:
In current Nextcloud versions, the browser call responsible for creating 404s in the trashbin is
`GET /apps/files_trashbin/preview?fileId=[ID]&x=32&y=32&mimeFallback=true&a=0`.
There is no `file` http argument anymore, so the whitelist will not catch those false positives.


ThomasCr:
that is already fixed in https://github.com/crowdsecurity/hub/pull/1087

https://github.com/crowdsecurity/hub/commit/21e221b2287700f1aa8046ee551bb8263c7e9199#diff-f5121c75367616c574b583068758dea174dc52dd0d772a4f957c2535784a76baR15

PR #898, opened 2023-12-27
Add waf check fp tool

AlteredCoder:
- Add a tool to run openappsec legitimate dataset against CrowdSec AppSec
- Add GitHub workflow to run it  (in parallel)
- Add concurrency on appsec-test, hubtest, and this workflow to avoid running a workflow multiple times simultaneously for the same PR.


waf-check/main.go
```diff
@@ -0,0 +1,69 @@
...
+		filesList, err = DownloadDataset(config.DatasetFolder)
+		if err != nil {
+			log.Fatalf("fail downloading dataset: %s", err)
+		}
+	} else {
+		filesList, err = listJSONFiles(config.DatasetFolder)
+	}
+
+	if len(filesList) == 0 {
+		fmt.Printf("No files in dataset '%s', please download with '-download'\n", config.DatasetFolder)
```
buixor, line 30: should we fatal on this to make sure we fail CI ?

waf-check/manager.go
```diff
@@ -0,0 +1,129 @@
...
+
+		resp, err := client.Do(&req)
+		if err != nil {
+			log.Fatalf("error doing request from file '%s' to '%s': %s", file, request.FullURL, err)
+		}
+
+		if resp.StatusCode == http.StatusForbidden {
+			responseBody, err := io.ReadAll(resp.Body)
+			if err != nil {
+				fmt.Printf("error reading response body for URL %s: %s\n", request.FullURL, err)
```
buixor, line 72: same as above, as it's for CI purpose, shouldn't we fail on this ?

buixor (changes requested):

buixor (approved):

PR #906, opened 2024-01-05
fix configserver-lfd tests

mmetc:


.tests/csf-lfd-logs/config.yaml
```diff
@@ -1,8 +1,8 @@
 parsers:
 - crowdsecurity/syslog-logs
-- crowdsecurity/configserver-lfd-logs.yaml
+- crowdsecurity/configserver-lfd-logs
```
AlteredCoder, line 3: i think it should be `crowdsecurity/s01-parse/configserver-lfd-logs.yaml`

AlteredCoder (changes requested):

.tests/csf-lfd-logs/config.yaml
rr404: sorry i had renammed csf-lfd-logs to configserver-lfd-logs also, so it also gonna mess the asserts

rr404: maybe

mmetc: yes I figured, thanks

PR #914, opened 2024-01-13
Add Miniflux collection

jbowdre:
Adds a collection, parser, and scenario for [Miniflux](https://github.com/miniflux/v2).


buixor (approved):
thanks for the contribution

Issue #814, opened 2023-08-28
overseerr


sahara101:
Hi all :)

I noticed we have collections for jelyseerr and oombi, would it be possible to add one for overseer? 
Would Plex also be possible?

Thanks!

LaurenceJJones:
Yes, if you can provide log samples and what you want to detect we can see about supporting it

sahara101:
I asked what could be checked on the overseerr discord, as I am not sure.

LePresidente:
i'll have a look at this shortly, since i just moved from jellyseerr to overseerr

andreasbrett:
Overseerr will produce the following log output in `overseerr.log`. Unfortunately when run in a Docker container the logged IP is from the Docker network subnet.

```
2023-12-10T15:17:39.030Z [warn][API]: Failed sign-in attempt using invalid Overseerr password {"ip":"::ffff:10.9.100.9","email":"someone@nonexisting.com"}
```

Plex is authenticating agains app.plex.tv so you won't see the logs. Then again I'd assume the Plex cloud infra to have a stronger security posture than our home labs.

PR #924, opened 2024-01-18
Create Overseerr collection based off jellyseerr collection.

LePresidente:
Created Overseerr collection based off jellyseerr collection.

they currently are technically two separate products, and this should lead to less confusion that the jellyseerr collection would work with it.

This will also complete
https://github.com/crowdsecurity/hub/issues/814


LaurenceJJones (approved):
Lgtm

PR #942, opened 2024-01-29
Adding WAF Coraza+Caddy parser/scenario

Barnoux:
Hello,

Hope this parser will find some love. 
The goal of this parser is to parse waf alert from coraza when coraza is integreted as a plugin in caddy.
the scenario is triggered based on the treshold of the inbound anomaly score setup by the users in the crs-setup.conf file.

The work on the parser is based on the work done by https://github.com/crowdsecurity/hub/blob/master/parsers/s01-parse/crowdsecurity/modsecurity.yaml

It's time for me to eat a cake and take a nap.


scenarios/barnoux/coraza.yaml
```diff
@@ -0,0 +1,15 @@
+type: trigger
+name: barnoux/coraza
+description: "Web exploitation detected via coraza"
+filter: evt.Meta.log_type == 'modsecurity' && evt.Parsed.ruleid == '949110'
```
LaurenceJJones, line 4: rather being coraza maybe should be named something to do with CRS?

LaurenceJJones:
Hey 👋🏻 Thank you for opening a PR!

We going to be need some tests for the parsers and scenarios. I left an initial comment since coraza is a modsecurity implementation meaning a scenario on the rule id might not be best since you dont have to use CRS.

scenarios/barnoux/coraza.yaml
Barnoux: yes you right, i will try to find a better name.

Barnoux:
> Hey 👋🏻 Thank you for opening a PR!
> 
> We going to be need some tests for the parsers and scenarios. I left an initial comment since coraza is a modsecurity implementation meaning a scenario on the rule id might not be best since you dont have to use CRS.

What type of input do you need for testing ? The rule id chosed in the scenario is based on the inbound anomaly score that is triggered and can be tuned by the user (https://coreruleset.org/docs/concepts/anomaly_scoring/#anomaly-score-thresholds). I did't find a better way to handle this.

scenarios/barnoux/crs-anomaly-score.yaml
```diff
@@ -0,0 +1,15 @@
+type: trigger
+name: barnoux/crs-anomaly-score
+description: "Web exploitation detected via Core Rule Set inbound anomaly scoring set by the user in crs-setup.conf"
+filter: evt.Meta.log_type == 'modsecurity' && evt.Parsed.ruleid == '949110'
+blackhole: 2m
```
LaurenceJJones, line 5: If you have a blackhole key it be best to also have a groupby since a single trigger will blackhole the whole bucket

LaurenceJJones (changes requested):

scenarios/barnoux/crs-anomaly-score.yaml
Barnoux: is it ok now ?

LaurenceJJones:
Okay, the last thing is adding a `collection` of

Parser
Scenarios (could include our standard modsec one 🤷🏻 )

Barnoux:
> Okay, the last thing is adding a `collection` of
> 
> Parser Scenarios (could include our standard modsec one 🤷🏻 )

I add the collection, is it looking good ?

ubergeek77:
Hey there, just passing by. I came across this PR as it's exactly what I'm looking for.

@LaurenceJJones , hope you don't mind the ping after so long, but are any more changes needed to get this merged? :slightly_smiling_face:

Barnoux:
> The only issue I have with is the scenario as its close to the original one we have for modsecurity. However, I did forget about this, so ill do this now.

What can i do to help you on this one ?

LaurenceJJones:
I will look into it, the only thing I dont like is the modification that was needed to the original caddy parser.

Maybe we can move this too `s02` level instead cause if at any point they change the log handler name it will break the chain. Moving to `s02` means we can still keep the original filter and then let the natural chain processing happen. The only thing is we have to execute before the original enrichers but this can be looked into.

Barnoux:
Hello, how can i help to close this MR ?

LaurenceJJones:
> Hello, how can i help to close this MR ?

Apologies it was on me, I decided just to merge as is, saves the back and forth plus me forgetting to stay on top of this.

If anybody has any issues, please open an issue and link logs but thank you @Barnoux , congratz on a valuable addition to the hub.

PR #955, opened 2024-02-06
[iptables/pf/windows] Add dst port as context

LaurenceJJones:
It interesting for users to know what ports were scanned for alert context


LaurenceJJones:
image

Should we prepend to proto? `tcp:1337` @buixor @mazzma12

Issue #958, opened 2024-02-07
Grafana logs are not parsed correctly when using LePresidente/grafana


LePresidente:
Reported on Docker

Grafana has updated there log format so new logs are not parsed,

```
logger=authn.service t=2024-02-06T07:24:56.502308134Z level=info msg="Failed to authenticate request" client=auth.client.form error="[password-auth.failed] failed to authenticate identity: [password-auth.invalid] invalid password"
logger=context userId=0 orgId=0 uname= t=2024-02-06T07:24:56.503804307Z level=info msg=Unauthorized error="[password-auth.failed] failed to authenticate identity: [password-auth.invalid] invalid password" remote_addr=X.XX.XXX.XX traceID=
```

The Parser should also be updated to support there Json format/docker . Will look into this

LePresidente:
The above has been done, grafana enterprise json logging only reports back a 401 error so it was not included

PR #959, opened 2024-02-07
[ WIP ] Added new grafana log format to parser.

LePresidente:
Grafana changed there log format, so this adds support for that
Also changed IP's in logs to be more visible 

This fixes https://github.com/crowdsecurity/hub/issues/958


parsers/s01-parse/LePresidente/grafana-logs.yaml
```diff
@@ -1,16 +1,28 @@
...
       apply_on: message
       statics:
         - meta: log_type
           value: grafana_failed_auth
-
+  - filter: UnmarshalJSON(evt.Parsed.message, evt.Unmarshaled, "grafana") in ["", nil]
+    nodes:
+    - filter: evt.Unmarshaled.grafana.msg == 'Invalid username or password'
+      nodes:
+        - grok:
```
LaurenceJJones, line 17: Do you want to use GROK to make sure they are compliant with the pattern?

Because if that is not the case you can just

```
      statics:
        - meta: log_type
          value: grafana_failed_auth
        - parsed: remote_ip
          expression: evt.Unmarshaled.grafana.remote_addr
        - parsed: timestamp
          expression: evt.Unmarshaled.grafana.t
```

parsers/s01-parse/LePresidente/grafana-logs.yaml
```diff
@@ -1,16 +1,28 @@
 onsuccess: next_stage
-#debug: false
+debug: true
```
LaurenceJJones, line 2: guessing left over from debugging, can we set `false` or comment it out

LePresidente:
Thanks yea that is a lot cleaner. all issues should be resolved now

LaurenceJJones (approved):
LGTM

PR #967, opened 2024-02-10
add NPMplus log parser and collection

Zoey2936:


LaurenceJJones:
Do we need to have an additional parser, or could the program filter be modified to match both since under the hood it just NPM

Zoey2936:
I use a different log format, so I think not

LaurenceJJones:
> I use a different log format, so I think not

Awesome, I couldn't tell from just the yaml, so yeah, a different parser is needed

.tests/npmplus-logs/parser.assert
```diff
@@ -0,0 +1,174 @@
+results["s01-parse"]["ZoeyVid/npmplus-logs"][0].Success == true
```
LaurenceJJones, line 1: Within the tests, can we have the full output, including all stages

LaurenceJJones (changes requested):

.tests/npmplus-logs/parser.assert
Zoey2936: like now?

LaurenceJJones: Yes!

LaurenceJJones:
I will review it on Monday once back in the office

Zoey2936:
Thanks!

LaurenceJJones (approved):
LGTM

PR #975, opened 2024-02-15
Adding QUIC.cloud CDN to the postoverflow scenarios

litespeedlars:
Adding QUIC.cloud, Litespeeds main own CDN, as a new post-overflow scenario for CDN whitelists.


LaurenceJJones (approved):
LGTM

PR #980, opened 2024-02-20
Added nginx error log filter and scenario for the ngx_http_limit_conn_module

Y3ll0w:
Updated the existing nginx parser by adding a filter for parsing the "connections limit exceeded" nginx error log messages.  

Create a new scenario `nginx-conn-limit-exceeded` based on the existing `crowdsecurity/nginx-req-limit-exceeded` scenario.
  
Added `.test` files and data for the new scenario.  

https://nginx.org/en/docs/http/ngx_http_limit_conn_module.html


Y3ll0w:
Missed something when updating the name from `con-` to `conn-`. All checks should pass now.

PR #997, opened 2024-03-11
add support for amavis

buixor:
- [amavis](https://amavis.org/) parser + scenario to ban IPs sending blocked/infected emails
 - associated collection

Pending questions:
 - Not sure about the `journalctl` filter for amavis
 - Not sure about the default file path for amavis
 - We could use `amavis_relay_ip` or `amavis_amavis_origin_ip` as the "bad ip". I went with `amavis_relay_ip` (as a malevolent sender could otherwise provide fake "origin ip").


LaurenceJJones (approved):
LGTM the only thing is we should add markdown for parsers and scenario even if something short like scenaro detects X in this case an infected email attempt

PR #998, opened 2024-03-12
Authelia update CLF and JSON

LaurenceJJones:
General update to authelia parser

@LePresidente hope you can have a look


LePresidente:
LGTM

PR #1017, opened 2024-04-06
Add ssh auth failure message variant for preauth without user

kyuuk:
ssh preauth may or may not have user part & also the `[perauth]` part

PS. i was not able to run the tests the documentation [here ](https://doc.crowdsec.net/docs/contributing/contributing_test_env/) seems to not work 

feel free to edit this PR or point me in the right direction to test my changes :)


LaurenceJJones:
fixed by:  #1168

kyuuk:
@LaurenceJJones not exactly the same, and this will still be missing entries when [`[preauth]` is not present at the end ](https://github.com/crowdsecurity/hub/pull/1168/files#diff-c7fd2e9adc0663b3d0d877e794c6910592e6c8c027001d87664e132434e876bbR15)

PR #1026, opened 2024-04-22
fix: disable debug log for wordpress scan scenario

GNU-Plus-Windows-User:


blotus:
Hello,

Thanks for the PR, I've actually just fixed this (and 2 other debug leftovers) in https://github.com/crowdsecurity/hub/pull/1029, so I'm going to close it.

PR #1054, opened 2024-06-12
Added a parser, scenario, and collection for Foundry VTT servers

eastcw:
Added a parser, brute force scenario, and collection for [Foundry VTT](https://foundryvtt.com/) servers.


parsers/s01-parse/eastcw/foundryvtt-logs.yaml
```diff
@@ -0,0 +1,33 @@
+name: eastcw/foundryvtt-logs
+description: "Parse Foundry VTT logs"
+filter: "evt.Parsed.program == 'foundryvtt'"
+debug: true
```
LaurenceJJones, line 4: debug statements should be false or commented

parsers/s01-parse/eastcw/foundryvtt-logs.yaml
```diff
@@ -0,0 +1,33 @@
...
+onsuccess: next_stage
+pattern_syntax:
+  DATE_YMD: "%{YEAR:year}-%{MONTHNUM:month}-%{MONTHDAY:day}"
+
+nodes:
```
LaurenceJJones, line 9: For the grok parsing since its all JSON maybe we should use the inbuilt https://docs.crowdsec.net/docs/next/expr/json_helpers#unmarshaljsonjsonblob-string-out-mapstringinterface-targetkey-string

Then it doesnt matter if the json format changes since we parse the keys to an object string

scenarios/eastcw/foundryvtt-bf.yaml
LaurenceJJones, line 1: Maybe we should add a slow variant?

LaurenceJJones (commented):
commenting for now: thank you for taking the time to create this collection 🦙 👍🏻

scenarios/eastcw/foundryvtt-bf.yaml
eastcw: Forgive me, I'm very much still learning how this works. Can you clarify what you mean? Do you know of any example configs similar to what you're thinking of already in the hub that I can look at?

LaurenceJJones: So when dealing with bruteforce you will battle with "fast" attacks and "slow" attacks typically slow attacks try to defeat detection by going just slow enough not to raise detections.

So for these scenario you would raise the leaktimer to something like `60s` or even `120s`, however, this causes a tradeoff between people generally forgetting their passwords so to counter this you raise the total capacity higher for example `10`.

You can find examples on our ssh bruteforce example https://app.crowdsec.net/hub/author/crowdsecurity/configurations/ssh-slow-bf

Also no need to say forgive me, we here to help and if you need anything let us know!

parsers/s01-parse/eastcw/foundryvtt-logs.yaml
eastcw: Good note. Is there a way to test a setup that uses UnmarshalJSON() ? Trying to use hubtest on any yaml that calls it throws a fatal error, even parsers that I know should be valid like crowdsecurity/laurel-logs

eastcw:
I've had a look and hopefully made the changes you requested. I've also included a suggested whitelist in the collection readme as I've noticed Foundry pretty regularly triggers the http-crawl-non_statics. I haven't included a whitelist config in the collection proper as it currently whitelists the whole Foundry subdomain and so needs to be customised for each installation. I'd be open to suggestions if there's a better way to structure it that would allow us to put a whitelist yaml in the collection.

PR #1093, opened 2024-08-20
Update parser to look at both sshd and sshd-session log lines

hobeone:
On Debian SID sshd-session shows up in the auth.log files with most of the messages that Crowdsec is looking for.  The parser was filtering to just sshd so was missing most of them.  This pull changes the filter to look at both sshd and sshd-session log lines.


LaurenceJJones:
Hey 👋🏻 

Can you regenerate the test files since you added a new line, it will need to output the additional lines.

hobeone:
Took a shot at this.

I think I needed to add a parser assertion to check that sshd and sshd-session are parsed correctly.  I'm not sure what I would check in the scenario.assert file.  Please let me know if there's a better way of doing this.

LaurenceJJones (approved):

PR #1096, opened 2024-08-22
Update pterodactyl-wings-logs parser

Lourys:
Update the parser to make sure it works with the latest version of pterodactyl-wings software

### DIFF
--- a/parsers/s01-parse/lourys/pterodactyl-wings-logs.yaml
+++ b/parsers/s01-parse/lourys/pterodactyl-wings-logs.yaml
@@ -1,17 +1,19 @@
 name: lourys/pterodactyl-wings-logs
 description: "Parse Pterodactyl wings logs"
+pattern_syntax:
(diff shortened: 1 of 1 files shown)


LaurenceJJones:
Hey 👋🏻 

Could we add the supported versions to the markdown file of the collection or parser so users know they need to be on a minimum version?

Lourys:
Hi!
I've just made changes so that all versions are supported. As the method isn't very important, I've replaced it with a regex that matches everything.

LaurenceJJones:
I will fix this and get the tests passing

LaurenceJJones:
Dont suppose you have some updated logs lines that we can add to tests?

Lourys:
Not really,
Here's a line with the method field if it can be useful:
`WARN: [Aug 22 08:32:21.074] failed to validate user credentials (invalid username or password) ip=10.87.200.23:52205 method=password subsystem=sftp username=rop0glbf.1f6a0e72`

PR #1098, opened 2024-08-25
Update pterodactyl-wings-logs parser

Zarklord:
My pterodactyl wings logs have a space character in front of WARN, add support for whitespace to appear before WARN and it can still properly parse the logfile.


LaurenceJJones (approved):
lgtm, but there shouldnt be any whitespace so many its a syslog thing but it doesnt break the actual parser itself 🤷🏻

Zarklord:
> lgtm, but there shouldnt be any whitespace so many its a syslog thing but it doesnt break the actual parser itself 🤷🏻

I'm just using the pterodactyl wings docker image, and mounting the log file into crowdsec, I took a look at the source code, and fhe prefix has been ` WARN` for the past several years from what I saw. 
https://github.com/pterodactyl/wings/blob/d7399489891f15ce85d9b1b8b93a6423080247ff/loggers/cli/cli.go#L27

Issue #1102, opened 2024-08-27
False positive: Nextcloud auto uploading from mobile clients (nextcloud behind nginx proxy)


brknkfr:
**Describe the bug**
Mobile nextcloud clients, authenticated with an app token (password token, get banned for checking existing files via `HEAD`.

**To Reproduce**
Install nextcloud behind nginx proxy. Connect nextcloud client via app token. Enable auto uploading of files (images for example) on nextcloud client on mobile device (Android, iPhone). The client gets banned after several tries with lots of following notices:

```
cscli alerts inspect --id 2079
- Date: 2024-08-27 13:19:43 +0200 +0200
╭─────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│       Key       │                                                        Value                                                        │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ datasource_path │ /var/log/nginx/access.log                                                                                           │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ datasource_type │ file                                                                                                                │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ http_args_len   │ 0                                                                                                                   │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ http_path       │ /remote.php/dav/files/**REDACTED**│
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ http_status     │ 403                                                                                                                 │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ http_user_agent │ Mozilla/5.0 (Android) Nextcloud-android/3.29.2                                                                      │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ http_verb       │ HEAD                                                                                                                │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ log_type        │ http_access-log                                                                                                     │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ service         │ http                                                                                                                │
├─────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ source_ip       │ 85.195.238.45                                                                                                       │
[...]

PR #1103, opened 2024-08-27
Fix for https://github.com/crowdsecurity/hub/issues/1102

brknkfr:
Fix for https://github.com/crowdsecurity/hub/issues/1102


LaurenceJJones:
Hey 👋🏻 

The only issue I see at the moment is `evt.Meta.http_path contains '/remote.php/dav/files/'` is similar to a line we have above and is on the similar lines.

I would only expect `403` response codes when the token you have is expired but maybe im abit naive on how the plugin /  app works in nextcloud.

I looked at your issue and maybe the `HEAD` request `403` is because the IP had a decision rather than the application itself?

The rest seem fine like `404` on `GET` but that can also be included in the original whitelisted line. 

Let me know your thoughts and if im abit naive on some parts 🦙

PR #1108, opened 2024-09-02
Update geoip-enrich.yaml to include default k3s ip range

abelfodil:
k3s flannel CIDR has the following ranges by default: https://docs.k3s.io/cli/server#networking

This will prevent spammy logs for k3s clusters


LaurenceJJones:
This should already be covered in `10.0.0.0/8` so this change is not needed.

Issue #988, opened 2024-03-03
Vaultwarden parser doesn't handle time zones


dlprows:
**Description**

Dominic-Wagner/vaultwarden parser doesn't handle time zones.

The pattern for parsing the timestamp `[%{DATE_YMD:date} %{TIME:time}\]` assumes that the logs will be written in UTC. But when properly configured, vaultwarden's logs are in local time.

When crowdsec parses the logs, it assumes UTC, and makes decisions accordingly. In my case, the decisions had the ban period entirely in the past.

This can be handled by changing the pattern to use `TIMESTAMP_ISO8601` and changing the vaultwarden log format to add `%z`

**Expected behavior**
Update the parser to handle timestamps. A recommended route is provided in the description.
Update the readme/setup directions to indicate the change needed in vaultwarden's admin to make the format have timezone.

LaurenceJJones:
Could you provide an example log line so we can test and ultimately add it to the test files so we don't break it moving forward

dlprows:
These are all the logs that the parser looks for

```
[2024-03-02 17:38:19.023-0700][vaultwarden::api::identity][ERROR] Username or password is incorrect. Try again. IP: 31.222.254.228. Username: blah@gmail.com.
[2024-03-02 17:42:20.407-0700][vaultwarden::api::identity][ERROR] Username or password is incorrect. Try again. IP: 31.222.254.228. Username: blah@gmail.com.
[2024-03-07 21:16:33.743-0700][vaultwarden::api::core::two_factor::authenticator][ERROR] Invalid TOTP code! Server time: 2024-03-08 04:16:33 UTC IP: 192.168.43.14
[2024-03-07 21:19:30.450-0700][vaultwarden::api::admin][ERROR] Invalid admin token. IP: 192.168.41.1
```


The default format in vaultwarden `Vaultwarden Admin Panel -> Advanced Settings -> Log timestamp format`
is
`%Y-%m-%d %H:%M:%S.%3f`

which would result in identical logs but without the -0700 in the timestamp

eg
```
[2024-03-02 17:38:19.023][vaultwarden::api::identity][ERROR] Username or password is incorrect. Try again. IP: 31.222.254.228. Username: blah@gmail.com.
[2024-03-02 17:42:20.407][vaultwarden::api::identity][ERROR] Username or password is incorrect. Try again. IP: 31.222.254.228. Username: blah@gmail.com.
[...]

PR #1117, opened 2024-09-18
enhance: Fix vaultwarden logs to allow non timezoned timestamps

LaurenceJJones:
Fix: #988

Use `TIMESTAMP_ISO8601` instead of a custom pattern as it handles a non timezoned string as well.


LaurenceJJones:
Update: we are waiting to merge this as we need to test what will happen in "live" mode when the log lines contains the offset, as this shouldnt use this timestamp as the timestamp is only for "cold" mode so we need to investigate the original issue.

PR #1122, opened 2024-09-26
Fix spelling mistake in http-dos-switching-ua scenario name

saz:


LaurenceJJones:
Hey 👋🏻 

Just note we are not ignoring this PR, the issue is currently if you change the name property the hub currently doesn't know this is the same scenario under a different name so people could download the scenario twice.

We are investigating ways to improve this but we cannot merge this PR till it is resolved.

PR #1124, opened 2024-10-01
enhance: support tls protocol in asterisk logs

crazybolillo:
Besides TCP and UDP, asterisk also supports TLS connections, logs for connections using said protocol were being ignored which enabled brute force attacks on Asterisk servers using TLS.


LaurenceJJones:
Wait for test to complete on the workflow so but LGTM to merge as soon as they pass!

LaurenceJJones (approved):

PR #1141, opened 2024-10-22
Add fortinet collection

AlteredCoder:
- Add fortinet parser 
- Add fortinet scenario to detect bruteforce on fortinet VPN
- Add fortinet alert context


buixor (approved):

LaurenceJJones:
> How do you get the logs into crowdsec?

Fortinet has a way to output to syslog endpoint, you can either use crowdsec syslog endpoint or the server syslog itself.

PR #1166, opened 2024-11-18
add mjs to list of static extensions

### LINKED ISSUE #1165: False Positive: mjs static files are counted as dynamic assets
mjs isn't counted as an static file and is causing false positives with applications that make heavy use of mjs files (Such as Nextcloud). I don't know where static resources are defined so I'm opening an issue here.

GNU-Plus-Windows-User:
closes #1165


LaurenceJJones (approved):
LGTM, we having issue with our remote sync to hub, so will wait till that is fixed then merge

Issue #1132, opened 2024-10-13
Crowdsec bans the wrong IP when OpenAppSec is behind a reverse proxy


alnviana:
I have the following structure:
Internet -> Reverse Proxy 1 -> OpenAppSec -> App1, App2, etc

The first proxy is configured to fill and pass the X-Forwarding-For, so it's value is "Real IP, Proxy 1 IP".
I configured OpenAppSec Source Identity to use this header instead of using Source IP.
I did a SQL Injection test, a event was triggered:
```json
{
  "eventData": {
    "proxyIP": "<REAL IP>",
    "sourceIP": "<PROXY 1 IP>",
    "httpSourceId": "<REAL IP>",
  }
}
```
Crowdsec reads the logs, parses the sourceIP and bans Proxy 1 IP.

Adding the Proxy 1's IP as a hop on Source Identity doesn't help, resulting in:
```json
{
  "eventData": {
    "sourceIP": "<PROXY 1 IP>",
    "httpSourceId": "<PROXY 1 IP>",
  }
}
```

With OpenAppSec directly exposed (Without Proxy 1) the Real IP appears as sourceIP, so it is banned correctly.

So I'm thinking of the following possible reasons for it not working:
- The parser was planned to work just when OpenAppSec is directly exposed to the internet.
  - If this is the case, can someone tell me the correct way to configure it? I know I can create a parser manually, but since it exists and is a prerequisite for collection, I think I need to do something else to avoid banning duplicates with the wrong IP.
- OpenAppSec is filling in the fields incorrectly.
  - I tried to find the documentation for these fields, what they should contain and how they should be used, but I couldn't find it.
  - It's strange that the Real IP is placed in the proxyIP field, after all, this is the IP of the source and not the Proxy, but that depends on your point of view. So, I think the name of the proxyIP field might not be the best, or at least it should have documentation explaining it. (I know, it's not up to you)
- The parser is using the wrong field.
  - It could be changed to use the proxyIP when it exists. Or maybe it shouldn't use sourceIP but httpSourceId... But as I said above, I couldn't find any information about these fields.

If anyone can help me, I'd be grateful. :)

alnviana:
Thanks for the quick reply. :)
I agree with you that it is good to plan this change well, as it could cause unwanted effects on those who are already using it. This care is the basics when it comes to security and resilience, which makes me like the project even more.

If you need any more information or help testing something, feel free to ask me.

chkp-sergeysh:
[...]

PR #1181, opened 2024-11-28
fix(parsers/openappsec): update real sourceIP field

he2ss:


rr404 (approved):
LGTM 
Should fix https://github.com/crowdsecurity/hub/issues/1132

PR #1194, opened 2024-12-18
Import SigmaHQ auditd rules

buixor:
- add `sigmahq` auditd rules
 - update the auditd and laurel parsers, reflect changes on associated scenarios


parsers/s01-parse/crowdsecurity/laurel-logs.yaml
```diff
@@ -17,7 +17,8 @@ nodes:
           - meta: service
             value: laurel
           - meta: log_type
-            expression: evt.Unmarshaled.laurel.SYSCALL.SYSCALL
+            expression: |
+              evt.Unmarshaled.laurel.SYSCALL.SYSCALL == "execve" ? "auditd_syscall_execve" : "auditd_syscall"
```
LaurenceJJones, line 21: Maybe we should just return

`'auditd_syscall_' + evt.Unmarshaled.laurel.SYSCALL.SYSCALL` ?

LaurenceJJones (approved):

PR #1208, opened 2024-12-31
'mkindex' to eventually replace go code; check from CI that output is the same

mmetc:


.index.json
```diff
@@ -16219,9 +16219,6 @@
...
    "description": "Detect exploit attempts via emerging threat rules",
    "author": "crowdsecurity",
-   "references": [
```
LaurenceJJones, line 16222: hmmm? I dont see how this was in the index to begin with but it not longer apart of the scenario so this good?

mmetc: yes, that's one of the fixed bugs.

LaurenceJJones (approved):

PR #1236, opened 2025-01-29
Add CVE-2024-6205

AlteredCoder:


appsec-rules/crowdsecurity/vpatch-CVE-2024-6205.yaml
```diff
@@ -0,0 +1,25 @@
+
+name: crowdsecurity/vpatch-CVE-2024-6205
+description: "PayPlus Payment Gateway WordPress plugin - SQL Injection (CVE-2024-6205)"
+rules:
+  - and:
```
seemanne, line 5: maybe also restrict to urls where  wc-api is payplus_gateway? Don't know how fp prone libinjection is, would put us on the safe side.

seemanne (approved):

PR #1244, opened 2025-02-06
Bind9 on Debian adds extra info

Olm666:
Example :
06-Feb-2025 10:32:43.895 security: info: client @0x7fcb602f4168 100.100.100.101#44717 (wr_test): query (cache) 'wr_test/A/IN' denied (allow-query-cache did not match)


LaurenceJJones:
Hey 👋🏻 

Merged as it didnt impact other tests and I added your log line minus the WAN, so we can ensure we dont break it in the future!

🦙

Olm666:
:ok_hand:

PR #1255, opened 2025-02-25
Optionally capture single quotes for login failed event

floatchallenged:
Login failed message sometimes uses single quotes (see [here](https://github.com/nextcloud/server/blob/b9aa3977e9c7b3363c5e327f1e3aefefc442669b/lib/private/User/Manager.php#L208) or [here](https://github.com/nextcloud/server/blob/b9aa3977e9c7b3363c5e327f1e3aefefc442669b/lib/private/User/Session.php#L439)).


LaurenceJJones:
Thanks for the PR it seems your changes dont align with the code examples you provided, even though they have two `'` marks, that means when the log is printed there only be a single `'`. So the changes should just wrap the elements with one `'?` on each side.

PR #1257, opened 2025-02-27
parser postfix: fix non-smtp-command for hosts which have a reverse dns

mschneider82:
currently "unknown" was hardcoded, unknown only matches if there is no reverse dns for the client ip.

### DIFF
--- a/parsers/s01-parse/crowdsecurity/postfix-logs.yaml
+++ b/parsers/s01-parse/crowdsecurity/postfix-logs.yaml
@@ -48,7 +48,7 @@ nodes:
           value: spam-attempt
   - grok:
       apply_on: message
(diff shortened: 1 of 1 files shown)


LaurenceJJones:
Good catch 👍🏻 

I looked at the my postfix logs and saw indeed there is a stretchoid scanner that didnt trigger it, so I added it as a test but of course its not.stretchoid.com 😉

LaurenceJJones (approved):
LGTM we wait for tests to pass then merge

mschneider82:
do we need to increase the version number?

LaurenceJJones:
> do we need to increase the version number?

Once we merge, the github action will generate it and increment it for us

mschneider82:
ok i maybe need to fix another rule in postfix, have to debug it first. it doesnt match SASL PLAIN/CRAM authentication very often. I have like 1000's in the log but just a few banned via the rule.

PR #1260, opened 2025-03-01
detect wordpress scanning of txt and md files

GNU-Plus-Windows-User:
Bots don't only scan the php files, they also scan for txt and md files and sometimes will only scan for one file type to try and evade detection.


LaurenceJJones:
Hey 👋🏻 

What do you think of only detecting `.md` and `.txt` files within `wp-content/plugins` only?

```
  evt.Meta.service == 'http' and
  evt.Meta.log_type in ['http_access-log', 'http_error-log'] and
  evt.Meta.http_status in ['404', '403'] and
  (
  Lower(evt.Meta.http_path) contains "/wp-" and
  Lower(evt.Meta.http_path) endsWith ".(php)"
  ) ||
  (
  Lower(evt.Meta.http_path) contains "/wp-content/plugins" and
  Lower(evt.Meta.http_path) matches "\\.(txt|md)$"
  )
```

Issue #1268, opened 2025-03-12
[LePresidente/authelia-logs] Not parsing missing user errors


AndreiArdelean1:
When an authentication attempt is made with a non-existing user, the message isn't parsed.

The error from authelia logs is:
```
time="2025-01-01T00:00:00+00:00" level=error msg="Error occurred getting details for user with username input 'MISSING_USER' which usually indicates they do not exist" error="user not found" method=POST 
```

Adding this parser to `nodes` solves the issue:
```
  ## Detect msg contains authentication attempt to parse out user info
path=/api/firstfactor remote_ip=192.168.0.1
  - grok:
      pattern: "Error .* getting details for user .* username input '%{AUTHELIA_USER:user}'.*" 
      expression: evt.Unmarshaled.authelia.msg
    statics:
      - meta: log_type
        value: auth_failed
```

LePresidente:
I'll add this today when i get a chance.

LePresidente:
Ok done, just need to wait for merge.

AndreiArdelean1:
Thank you!

LaurenceJJones:
Merged, please note as `1.6.6` recently you must be on this version to get the newer version of the parser. If you are within a container you will just need to wait whilst we finalize the container image 🦙

PR #1273, opened 2025-03-13
Updated Authelia Parser due to new log format for Unknown Users

LePresidente:
Fixes issue https://github.com/crowdsecurity/hub/issues/1268


LaurenceJJones:
will merge as soon as possible, we are releasing `1.6.6` today so once that is completed, I will merge.

PR #1293, opened 2025-03-23
Change scenario username meta to match with the parser username metaFix typo

LearningSpot:


LearningSpot:
> We can merge for now but maybe we might want to add `target_user` to meta if we want this collection to be used by context:
> 
> https://github.com/crowdsecurity/hub/blob/b9f56a06b7f8a24cf877a32b7eb3178e34ddb5c5/contexts/crowdsecurity/bf_base.yaml#L1-L4

That's great. I didn't think about that. Or I would just change both of them to target_user. So won't need to change later. Or do you want me change now and pr again?

PR #1371, opened 2025-05-29
Update GitLab parser to use UnmarshalJSON

timokoessler:
As suggested by @LaurenceJJones in #1367


LaurenceJJones:
Awesome was just about to jump on this, little suggestion move the unmarshaljson to the filter at the top next to the program check.

This ensures the groks are only focused on checking the properties.

Also if you can regenerate the parsers asserts as there will be new data to check.

PR #1393, opened 2025-06-12
ez setup

sabban:
easy setup scenarios with tests


buixor:
- Let's make sure that all payloads contain a human understandable string such as "CROWDSEC-TEST-..."
 - Add a out-of-band WAF scenario

sabban:
> * Let's make sure that all payloads contain a human understandable string such as "CROWDSEC-TEST-..."
> 
>     * Add a out-of-band WAF scenario

done :)

LaurenceJJones:
> > * Let's make sure that all payloads contain a human understandable string such as "CROWDSEC-TEST-..."
> >   
> >   * Add a out-of-band WAF scenario
> 
> done :)

You still need to add the rule to the collections of `vpatch` and `generic` as we install both as adding to appsec configuration is not enough to download it.

appsec-rules/crowdsecurity/appsec-generic-test.yaml
```diff
@@ -0,0 +1,25 @@
+name: crowdsecurity/appsec-generic-test
+description: "AppSec heartbeat – trigger on GET /crowdsec-test-NtktlJHV4TfBSK3wvlhiOBnl`"
```
buixor, line 2: maybe we should have a consistent naming `healthcheck` or something ? heartbeat isn't accurate I think

scenarios/crowdsecurity/appsec-generic-test.yaml
```diff
@@ -0,0 +1,14 @@
+type: trigger
+format: 3.0
+name: crowdsecurity/appsec-generic-test
+description: "generate an alert for appsec out of band rule for testing"
```
buixor, line 4: same as above

PR #1396, opened 2025-06-17
ez setup following tko's review

sabban:


tko:
@sabban I believe you mean to be pinging @timokoessler instead

timokoessler:
Thanks for mentioning me @tko, but this has nothing to do with me.

LaurenceJJones:
Let's clear up the confusion @buixor initials is tko so sorry for 2 false tags but pleasure meeting 😄

PR #1398, opened 2025-06-19
match on HEAD too

sabban:
match on HEAD too


buixor:
can we add it to the tests too plz ? <3

buixor (approved):

PR #1401, opened 2025-06-22
Bind9 parse logs without client id

nvtkaszpir:
Some requests may come without client id and thus they should be also processed properly


LaurenceJJones:
Hey 👋🏻 

Merged as it didnt cause any regression in tests, will be available to all users that are on 1.6.9 and will be live in about 5 minutes.

nvtkaszpir:
hm I just realized I have not bumped the related versions of the parser, because there is none....
Is there any action required because I think hub may thing there is no change?

LaurenceJJones:
> hm I just realized I have not bumped the related versions of the parser, because there is none.... Is there any action required because I think hub may thing there is no change?

The CI handles this for us: https://github.com/crowdsecurity/hub/commit/e821617ab1d14d89a77561180ecfa60e1b7e3acc

Issue #940, opened 2024-01-29
Unifi OS 3 aren't being parsed


GNU-Plus-Windows-User:
**Describe the bug**
Unifi OS 3 and newer logs are not being parsed correctly, resulting in detection scenarios such as port scanning not working correctly.

**To Reproduce**
1. Install the unifi collection ``cscli collections install crowdsecurity/unifi`` and reload crowdsec
2. Setup a syslog endpoint via acquis.yaml using the following yaml:
```
source: syslog
listen_addr: 0.0.0.0
listen_port: 514
labels:
 type: unifi
```
3. Configure a Unifi OS 3 console or newer to log to the syslog endpoint
4. check crowdsec logs and see ``time="03-11-2023 04:48:35" level=error msg="could not parse message: version must be 1" client=0.0.0.0 type=syslog``
5. check ``cscli metrics`` and see no logs are being parsed
```
╭────────────────────┬────────────┬──────────────┬────────────────┬────────────────────────╮
│       Source       │ Lines read │ Lines parsed │ Lines unparsed │ Lines poured to bucket │
├────────────────────┼────────────┼──────────────┼────────────────┼────────────────────────┤
│ syslog:0.0.0.0     │ 295        │ -            │ 295            │ -                      │
╰────────────────────┴────────────┴──────────────┴────────────────┴────────────────────────╯
```

**Expected behavior**
Logs should be parsed

**Screenshots**
N/A

**Additional context**
This issue was originally reported within the CrowdSec Discord

LaurenceJJones:
> check crowdsec logs and see time="03-11-2023 04:48:35" level=error msg="could not parse message: version must be 1" client=0.0.0.0 type=syslog

so the error is happening within syslog acquisition itself

https://github.com/crowdsecurity/crowdsec/blame/4e724f6c0a54ad1c67eeab6ca3be62f00ee0cf20/pkg/acquisition/modules/syslog/syslog.go#L209

It not even hitting the parsers at all..... so what format is it if its not `RFC3164` or `RFC5424`

GNU-Plus-Windows-User:
@LaurenceJJones I'm not sure what to look for, so let me know if you are missing some specific logs.

I didn't remove the MAC address from the last log line, that's how it was sent.
```
[...]

PR #1408, opened 2025-07-07
enhance: Attempt s00 unifi

LaurenceJJones:
s00 unifi to handle the firewall logs, formats to iptables parsers and uses the iptables sceanrio

Since unifi not longer does the old logs I have removed them and only uses the newest syslog logs.

#940


Speedliner:
any updates on that? I would really love to see that change to use cs-unifi-bouncer again with the newest UDM Pro firmware.

LaurenceJJones:
> any updates on that? I would really love to see that change to use cs-unifi-bouncer again with the newest UDM Pro firmware.

Problem is currently unifi is switching all logs from syslog to CEF, so we need some examples from people. 

Edit: also this parsers is not needed to use the bouncer, so if you have something hosted behind Unifi you can still use it but simply can't react on Unifi logs yet.

PintjesB:
> > any updates on that? I would really love to see that change to use cs-unifi-bouncer again with the newest UDM Pro firmware.
> 
> Problem is currently unifi is switching all logs from syslog to CEF, so we need some examples from people.
> 
> Edit: also this parsers is not needed to use the bouncer, so if you have something hosted behind Unifi you can still use it but simply can't react on Unifi logs yet.

My policy logs are still the same as before: (Network OS 9.3.45)
`Jul 30 05:41:41 UDMP-DTC [WAN_LAN-D-10001] DESCR="cs-unifi-bouncer-ipv4-External->Inter" IN=eth8 OUT=br202 MAC=74:ac:b9:1c:62:e5:00:17:10:2b:31:a9:08:00 SRC=45.55.158.168 DST=192.168.202.6 LEN=60 TOS=00 PREC=0x00 TTL=51 ID=9060 DF PROTO=TCP SPT=36051 DPT=4443 SEQ=1826324748 ACK=0 WINDOW=65535 SYN URGP=0 MARK=1a0000 `

PR #1418, opened 2025-07-18
Remove extra whitespaces

Jgigantino31:
Remove any extra unintentional whitespaces.


LaurenceJJones:
I dont believe we should merge this just based on whitespace, it doesnt actually cause an issue and will cause thousands of engines to get this update and restart for no benefit.

But I will leave it open to comment for other team members.

PR #1420, opened 2025-07-20
Add parser to allow all IPv6 address in a /64 to be combined into one shared bucket

Jgigantino31:
Add parser to allow all IPv6 addresses in a /64 to be combined into one shared bucket. This parser must be used with the crowdsecurity/ipv6_to_range postoverflow. Should there be an IPv6 collection that has both?

This is accomplished by using `TrimSuffix(IpToRange(evt.Meta.source_ip,"/64"),"/64")` to rewrite the source IP in the parser/s02-enrich stage to force the lower 64 bits to zero, which is still a valid IPv6 address. When going to scenarios designed for IP addresses all IPv6 addresses in a /64 will now go into a single bucket. On overflow, the crowdsecurity/ipv6_to_range postoverflow remediates on the entire /64 range.

Example effects on source_ip:

2001:db8:1234:5678::abcd => 2001:db8:1234:5678::  
2001:db8:1234:5678::1234 => 2001:db8:1234:5678::  
2001:db8:1234:5678::5678 => 2001:db8:1234:5678::  
2001:db8:1234:5678:abcd:1234:ef10:5678 => 2001:db8:1234:5678::  
2001:db8:1234:5678:4545:cdcd:6868:dada => 2001:db8:1234:5678::  
2001:db8:abcd:2020:abcd:1234:ef10:5678 => 2001:db8:abcd:2020::

image


.tests/ipv6-parser/parser.assert
```diff
@@ -0,0 +1 @@
+
```
blotus, line 1: Please fill the parser assert file (even though you have added assertions for the scenario to validate the behavior, it's still a parser, so we also want to test it by itself).

Jgigantino31: Understood, updated parser.assert file with test results.

Jgigantino31:
Anything stopping this from being merged?

PR #1426, opened 2025-07-25
Update Jellyfin and Jellyseerr whitelists and tests

Jgigantino31:
Jellyfin-whitelist test was missing a test for when /Sessions/Playing/Progress returns 403. Made whitelist itself more strict in how it matches.

Jellyseerr whitelist added in #1413 was missing a test, added here. Made whitelist itself more strict in how it matches.


buixor:
ty for your contribution!

PR #1437, opened 2025-08-05
add some more scenarios to the NPMplus collection by default

Zoey2936:
I also have a question, is it possible to rename ZoeyVid/npmplus to ZoeyVid/NPMplus or would this break exiting instances? Thanks!

### DIFF
--- a/collections/ZoeyVid/npmplus.yaml
+++ b/collections/ZoeyVid/npmplus.yaml
@@ -5,7 +5,10 @@ parsers:
 collections:
   - crowdsecurity/appsec-virtual-patching
   - crowdsecurity/appsec-generic-rules
+  - crowdsecurity/appsec-wordpress
   - crowdsecurity/base-http-scenarios
+  - crowdsecurity/http-cve
+  - crowdsecurity/wordpress
   - crowdsecurity/modsecurity
   - openappsec/openappsec
 scenarios:


LaurenceJJones:
> I also have a question, is it possible to rename ZoeyVid/npmplus to ZoeyVid/NPMplus or would this break exiting instances? Thanks!

The problem would be that existing users that have `collections` key defined in their compose.yaml would need to update the definition to the new caps alternative or else they will most likely crash on startup due to it not existing in the index anymore.

We started laying the ground work for the index to be more resilient on renaming / deprecation EG: if users defines `ZoeyVid/npmplus` it would know that it has been renamed to `ZoeyVid/NPMplus` but at the moment this is still in WIP.

Zoey2936:
> > I also have a question, is it possible to rename ZoeyVid/npmplus to ZoeyVid/NPMplus or would this break exiting instances? Thanks!
> 
> The problem would be that existing users that have `collections` key defined in their compose.yaml would need to update the definition to the new caps alternative or else they will most likely crash on startup due to it not existing in the index anymore.
> 
> We started laying the ground work for the index to be more resilient on renaming / deprecation EG: if users defines `ZoeyVid/npmplus` it would know that it has been renamed to `ZoeyVid/NPMplus` but at the moment this is still in WIP.

Ok, thanks! If it is possible in the future, it would be nice if it would be renamed. From my side it would not be a huge problem, if it breaks. Since NPMplus has often breaking changes, so the users need to read the changelogs in any case and since this collection is only used for NPMplus I could mention the rename in the NPMplus changelogs. But then the hub update and the NPMplus release need to be at the same time point. But as it is only a rename, it is just nice to have, not needed.

PR #1459, opened 2025-09-05
make OOB CRS appsec config generate alerts by default

buixor:
Improve a bit the explaination of OWASP CRS OOB collection, add the ability to create alerts by default


LaurenceJJones:
Just to ask, if you install and use the CRS collection this would mean **all** out of bands rules now send alerts?

Issue #1460, opened 2025-09-05
Obvious data center IPs are being classified as residential


GNU-Plus-Windows-User:
**Describe the bug**
Data center IP addresses such as AWS IPs are being marked as residential by CTI.

**To Reproduce**
https://app.crowdsec.net/cti/54.202.255.19

**Expected behavior**
AWS is not an residential ISP provider, it should be marked as an data center IP.

**Screenshots**
N/A

**Additional context**
N/A

mazzma12:
Hey, thank you for reporting. That's an issue we are aware of. This tag is currently experimental, and we will iterate over it this week. I will let you know once an update is made.

PR #1463, opened 2025-09-08
fix(classifications): value  mistyped for connection-type classif

mazzma12:


seemanne:
closes https://github.com/crowdsecurity/hub/issues/1460

Issue #1478, opened 2025-09-22
NPMplus parser incorrectly processes some user agent strings


Zoey2936:
See: https://github.com/ZoeyVid/NPMplus/issues/2091
I need some help on how to best fix this, see the Issue in the NPMplus repo for details

LaurenceJJones:
I read your ping but then went on annual leave, will look first thing tomorrow.

Zoey2936:
> I read your ping but then went on annual leave, will look first thing tomorrow.

Thanks. No preasure, take the time needed.

PR #1479, opened 2025-09-23
fix: npmplus does not log refferer or UA around double quotes

LaurenceJJones:
fix: #1478 

@Zoey2936 just confirm that in all cases http refferer is never wrapped in quotes and neither is useragent?


Zoey2936:
https://github.com/ZoeyVid/NPMplus/blob/19c0656dc7dba534882efabf09c3ae32f26d5aa8/rootfs/usr/local/nginx/conf/nginx.conf#L46
This is the used log format, so no it should be wrapped in quotes

PR #1564, opened 2025-11-03
Add support for EventLogs via Grafana Alloy

adamrushad:
## Description

* Updated windows-logs and windows-auth parsers to support the JSON format that Grafana Alloy transforms Windows Event Logs into before pushing them

* Updated tests for windows-logs and windows-bf to support JSON data

## Checklist

 - [ ] For new parsers or scenarios, tests have been added (N/A)
 - [ ] I have run the hub linter and no issues were reported (see contributing guide) (
 - [ ] AI was used to generate any/all content of this PR


parsers/s00-raw/crowdsecurity/windows-logs.yaml
```diff
@@ -1,24 +1,45 @@
...
+      - parsed: EventID
+        expression: string(int(evt.Unmarshaled.wineventlog.event_id))
+      - parsed: Source
+        expression: evt.Unmarshaled.wineventlog.source
+      - parsed: Computer
+        expression: evt.Unmarshaled.wineventlog.computer
+      - parsed: UserSID
+        expression: XMLGetNodeValue(evt.Unmarshaled.wineventlog.event_data, "/Data[@Name='UserID']")
+      - parsed: EventData
+        expression: evt.Unmarshaled.wineventlog.event_data
```
blotus: We could update `evt.Line.Raw` directly with the XML, that way, all subsequent parsers wouldn't have to check if the line was in JSON or XML:
```
      - target: evt.Line.Raw
        expression: evt.Unmarshaled.wineventlog.event_data
```

blotus (commented):
Sorry for the delay in reviewing this.

I don't like that we check the format of the log line in both parsers.
I've added a suggestion to overwrite `evt.Line.Raw` if the source is in JSON format, so that other parsers can just use the XML data, as if it was coming from the event log directly.

parsers/s00-raw/crowdsecurity/windows-logs.yaml
adamrushad: @blotus Sorry for the ping. I know you're busy, but I wanted to check back in on this to make sure the caveats above are acceptable. Specifically the notes about a couple of fields that I can't reproduce from the JSON, along with the caveats about future maintenance needs to map text back to numeric values.

PR #1577, opened 2025-11-06
fix: pf-logs IPv6 service detection using empty string check

LaurenceJJones:
## Description

pf-logs IPv6 service detection using empty string check

parsed variables are never nil so it was never setting ipv6 proto if it was available also it seems to be uppercase so wrapping in lower function to ensure it always lowercase.

## Checklist

 - [x] For new parsers or scenarios, tests have been added 


LaurenceJJones:
all tests are passing and fixes my regression so merging without approval

PR #1623, opened 2026-01-02
Remove whitelist-good-actors from linux.yaml

daikoz:
linux is a core collection: "core linux support : syslog+geoip+ssh". It should not contain crowdsecurity/whitelist-good-actors !!!

Moreover, whitelist-good-actors contain wrong actor which don't respect robots.txt.
And it added recently.

All whitelist should be add manually by server's administrator, not in core collection.

## Checklist

 - [ X] I have read the [contributing guide](https://docs.crowdsec.net/docs/next/contributing/contributing_hub)
 - [ X] I have tested my changes locally
 - [ X] For new parsers or scenarios, tests have been added 
 - [ X] I have run the hub linter and no issues were reported (see contributing guide)
 - [ X] Automated tests are passing
 - [ X] AI was used to generate any/all content of this PR

### DIFF
--- a/collections/crowdsecurity/linux.yaml
+++ b/collections/crowdsecurity/linux.yaml
@@ -4,7 +4,6 @@ parsers:
   - crowdsecurity/dateparse-enrich
 collections:
   - crowdsecurity/sshd
-  - crowdsecurity/whitelist-good-actors
 description: "core linux support : syslog+geoip+ssh"
 author: crowdsecurity
 tags:


LaurenceJJones:
Hey @daikoz thank you for opening a PR, i'll give some background on why we decided to make it within the standard linux collection.

Before hand we was getting about 100 - 200k signals per day based on these ranges, IPs. Because unfortunately users just install CrowdSec, go out to find other collection to install other than ones they _need_ cause they dont really understand they are banning ranges they can ultimately bring down their sites (cloudflare CDN gets banned, your site is unreachable for users going through cloudflare cdn).

Before we added this change we only had 600 instances that actively installed this VS the 200k instances that are out their.

> Moreover, whitelist-good-actors contain wrong actor which don't respect robots.txt.

If you can provide more content on who, what they did then we come be come around, but blanketing saying "administrators should install it" when we have the stats on our end to prove users just dont.

daikoz:
For example, as highlighted in this Datadome research on Facebook being used as a proxy by web scraping bots: [How Facebook Was Used as a Proxy by Web Scraping Bots](https://datadome.co/threat-research/how-facebook-was-used-as-a-proxy-by-web-scraping-bots/).

Applying a default whitelist to all users is inherently risky. It unnecessarily expands the attack surface for websites that do not rely on these third-party services, exposing them to potential threats such as DDoS or scraping attacks... A security tool should never enable such broad allowlists by default, as it compromises the principle of least privilege.

For this reason, we have removed this collection for all ours clients.

PR #1754, opened 2026-04-06
Fix typo

herrbischoff:
## Description

Fix "probbing" to "probing".

## Checklist

n/a


blotus:
Hello,

Thanks for your PR but unfortunately, we cannot accept PRs fixing typos in scenarios name: it could break existing configurations that rely on the name of a specific scenario (eg, a postoverflow or a custom profile)

herrbischoff:
Maybe I should have mentioned that using `cscli scenarios install crowdsecurity/http-xss-probbing` doas not work while `cscli scenarios install crowdsecurity/http-xss-probing` does. Just in case this makes any difference. If not, it's fine. I'm just in the habit of fixing things whenever I notice them instead of putting it off and inevitably forgetting about them.

PR #1762, opened 2026-04-10
feat: add OpenVPN parser and bruteforce detection

PrOOnOOb:
Adds a parser for OpenVPN logs supporting both syslog and ISO8601 timestamp formats, and a scenario to detect and ban IPs performing TLS bruteforce/probing attacks.


sabban:
Hi,

We already have something for OpenVPN brute force https://app.crowdsec.net/hub/author/pserranoa/scenarios/openvpn-bf. What would this bring differently?

Regards,

PrOOnOOb:
Hi @sabban,

There are a few meaningful differences that motivated this PR:

Parser approach:
The pserranoa parser filters on evt.Line.Labels.type == 'openvpn', which requires the user to manually set log_type: openvpn in their acquisition config and reads the log file directly. In practice, OpenVPN on most Linux systems logs via syslog/journald, so this PR is built around that typical setup — it uses crowdsecurity/syslog-logs as a pre-parser and identifies OpenVPN events by evt.Parsed.program == 'ovpn-server', which works out of the box without extra acquisition configuration.

Dual timestamp support:
The pserranoa parser only supports TIMESTAMP_ISO8601. This PR also handles legacy BSD syslog timestamps (Apr  5 11:26:07 ...), which are still common on older or minimal systems.

Pattern matching:
Because this parser builds on syslog-logs, patterns are applied only to the extracted message field rather than the full raw log line, making them cleaner and less brittle.

That said, I'm happy to discuss whether these should be merged into the existing pserranoa collection or kept separate. Let me know what makes more sense for the hub's structure.

sabban:
Hi, 

Sorry for the delayed answer.

I updated the pserranoa openvpn parser to not require evt.Line.Labels.type (this shouldn't have been accepted in the first place).

If you could work around the PR to use this parser (it's better to have only one parser that fits all requirements), that would be great.

Thanks !

Regards,

PrOOnOOb:
Hi @sabban, thanks for updating the parser  with the filter and syslog patterns now covering the out-of-the-box syslog/journald setup, this PR is no longer needed. Closing it. Thanks!

PR #1775, opened 2026-04-16
Backport change to 1.7.6 branch

blotus:
When 1.7.7 was released, version.crowdsec.net was not updated properly, meaning the 1.7.6 version still thought it was the latest available release and chose to keep using the `master` branch of the hub.

Before we can update version.crowdsec.net with the actual latest version, we need to backport anything that was pushed to the hub since the release of 1.7.7 to make sure that when 1.7.6 switches to the 1.7.6 branch, it does not lose any parsers/scenarios/... content.


sabban (approved):
lgtm

PR #1821, opened 2026-06-08
Guezli/postfix-honeypot-users: align doc wording with filter scope

Guezli:
## Description

Follow-up to #1806 addressing the two Copilot review comments that
remained unresolved at merge time:

- The description, top-of-file comment block and rendered `.md` all
  said "SASL LOGIN", but the filter only checks for `SASL` + `authentication failed`
  and therefore matches every SASL mechanism the upstream
  `crowdsecurity/postfix-logs` parser flags (LOGIN, PLAIN, CRAM-MD5, DIGEST-MD5).
  Wording is now "SASL authentication" so docs and filter agree.
  The broader scope is intentional -- catching a PLAIN attempt against
  `postmaster@` is just as much a clear attacker signal as LOGIN.
- The `.md` username list was missing `test@`, which the scenario regex
  includes. Added it.

## Functional impact

None. Pure doc/wording alignment. `cscli hubtest run postfix-honeypot-users --clean`
stays green locally, `hublint check` clean.

## AI assistance

Claude Code generated the wording-alignment edits based on the Copilot
review comments left on #1806.

### DIFF
--- a/scenarios/Guezli/postfix-honeypot-users.md
+++ b/scenarios/Guezli/postfix-honeypot-users.md
@@ -1,11 +1,15 @@
 ## Postfix SASL honeypot-username instant ban
 
-Instant-bans IPs that try postfix SASL LOGIN with a honeypot username --
-well-known admin / role addresses (`postmaster@`, `admin@`, `info@`,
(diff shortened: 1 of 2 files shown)


sabban (approved):
lgtm

PR #1825, opened 2026-06-15
do not make decision required to do RDNS enrichment

buixor:
- fixed a bug where RDNS based postoverflow whitelists wouldn't work because RDNS enrichment only happens **if** there is a decision in the alert.


sabban (approved):
lgtm

PR #1826, opened 2026-06-15
WAF Challenge Collection

buixor:


buixor:
in terms of `crowdsecurity/challenge-exclude-*` should we be more radical and exclude static resources that could lead to further issues ? (favico, images, `.ics` files, font files etc.)

PR #1833, opened 2026-06-23
feat: add PumbaLP/cowrie collection (parser & scenario)

PumbaLP:
## Description

This PR introduces the PumbaLP/cowrie collection, which bundles an aggressive detection scenario (cowrie-precision) and a dedicated parser for Cowrie Honeypot (SSH/Telnet) JSON logs.

Unlike standard SSH brute-force detection, this scenario targets automated environment discovery and post-exploitation inside the honeypot. It triggers a leaky bucket overflow immediately upon connection attempts (`cowrie-login`) or emulated command executions (`cowrie-command`), routing the offending IP directly to local remediation profiles.

Functional tests and a sample log file with an anonymized private IP have been added under `.tests/cowrie-precision/`.

## Checklist

 - [x] For new parsers or scenarios, tests have been added 


.tests/cowrie-precision/config.yaml
```diff
@@ -0,0 +1,6 @@
+parsers:
+- crowdsecurity/cowrie-logs
```
sabban, line 2: This loads crowdsecurity/cowrie-logs, while this PR adds and packages PumbaLP/cowrie-logs. As a result, the new JSON parser is not covered by the added test fixture, meaning CI could pass even if the parser shipped with the collection is broken or incompatible.

sabban (changes requested):
In addition to the review comments, this PR doesn't currently include any parser.assert or scenario.assert, which are needed for the tests. The initial checklist is there to help ensure these requirements are covered, so please make sure it accurately reflects the current state of the PR.

PumbaLP:
​@sabban I've added the missing assertion files (parser.assert and scenario.assert) and updated the test configuration. All local tests are passing now!

PR #1839, opened 2026-07-07
CI hardening

blotus:
- Pin actions SHA
 - Least privileges on workflows
 - No creds persistence
 - Remove 3rd party actions
 - Enable dependabot for actions


sabban (approved):
lgtm

PR #1882, opened 2026-09-07
WAF rule for CVE-2023-54391 (Proxmox VE authentication bypass)

mazzma12:
Virtual patch for CVE-2023-54391: POST to `/access/ticket` carrying a `tfa-challenge` that is not a server-issued `PVE:` ticket, or paired with a password that is not a typed second-factor response (`totp:`, `webauthn:`, `yubico:`, `recovery:`). Mirrors the alert-context rule crowdsecurity/alert-context-rules@39a8876 and the public nuclei template.
Test covers the exploit request (403) plus a legitimate 2FA second step and a plain login (pass through).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RVE7PF9rt4CgyE2vRH2nRW


mazzma12:
Closing: a rule here cannot reliably separate the bypass from a legitimate second-factor login, so we will not ship a virtual patch for this CVE.

PR #1886, opened 2026-09-09
sshd-logs: match max-auth-attempts and Disconnecting invalid user

### LINKED ISSUE #1880: `sshd-logs`: missing patterns for OpenSSH 9.6 "maximum authentication attempts exceeded" and "Disconnecting invalid user" messages
Two common OpenSSH log message formats produced by OpenSSH 9.6p1 are not matched by any grok pattern in `crowdsecurity/sshd-logs`, meaning genuine invalid-username / auth-exhaustion events are silently dropped at the `s01-parse` stage instead of being classified as `ssh_failed-auth`.
 
Confirmed via `cscli explain --type syslog -v` against freshly generated log lines - the raw syslog line parses correctly at `s00-raw` (`program: sshd` extracted correctly), but fails all `s01-parse` nodes including `sshd-logs`.
 
Example unmatched lines:
```
Sep  4 23:15:41 uptimekuma sshd[1714650]: error: maximum authentication attempts exceeded for invalid user nonexistentuser123 from 127.0.0.1 port 56876 ssh2 [preauth]
Sep  4 23:15:41 uptimekuma sshd[1714650]: Disconnecting invalid user nonexistentuser123 127.0.0.1 port 56876: Too many authentication failures [preauth]
```
 
Why existing patterns don't catch these:
- `SSHD_INVALID_USER` expects `Invalid user X from IP` - doesn't cover "error: maximum authentication attempts exceeded for invalid user...".
- `SSHD_PREAUTH_AUTHENTICATING_USER` expects `Connection (closed|reset) by...` - doesn't cover "Disconnecting invalid user..." (different verb, no "Connection").
- `SSHD_PREAUTH_AUTHENTICATING_USER_ALT` expects `Disconnected from (authenticating|invalid) user...` - close, but actual wording is "Disconnecting invalid user..." (different verb form, no "from").
Environment: OpenSSH `9.6p1 Ubuntu-3ubuntu13.19`, CrowdSec `1.8.1`, hub index up to date at time of testing (`cscli hub update` / `cscli parsers upgrade crowdsecurity/sshd-logs` both reported nothing to do).
 
## To Reproduce
 
1. On a host running OpenSSH 9.6+, trigger an invalid-user connection that exhausts auth attempts:
```
   ssh -o PreferredAuthentications=publickey nonexistentuser@target-host
```
   (repeated automatic key offers trigger a server-side "Too many authentication failures" disconnect)
2. Check `/var/log/auth.log` for the resulting `sshd` lines shown above.
3. Run `cscli explain --file <log-with-these-lines> --type syslog -v` and observe both lines fail at `s01-parse` under `sshd-logs`.
 
## Expected behavior
 
Both lines should be matched by `sshd-logs` and classified with `log_type: ssh_failed-auth`, with `target_user` and `sshd_client_ip` populated, so they can feed brute-force/invalid-user detection scenarios.
 
## Screenshots
 
N/A - text log output only, included above and in Additional context.
 
## Additional context
 
Suggested fix - add two new grok patterns to `sshd-logs.yaml`:
```yaml
SSHD_MAX_AUTH_ATTEMPTS: 'error: maximum authentication attempts exceeded for invalid user %{USERNAME:sshd_invalid_user} from %{IP_WORKAROUND:sshd_client_ip} port \d+ ssh2 \[preauth\]'
SSHD_DISCONNECTING_INVALID_USER: 'Disconnecting invalid user %{USERNAME:sshd_invalid_user} %{IP_WORKAROUND:sshd_client_ip} port \d+: Too many authentication failures \[preauth\]'
```
with corresponding `nodes:` entries following the same `log_type: ssh_failed-auth` / `target_user` static pattern already used elsewhere in the file.
 
Have a working local patch (a separate custom parser layered alongside the official one, same filter/stage) already validated against real traffic - happy to submit this as a PR if useful.

arpitjain099:
Closes #1880.

Two OpenSSH messages fell through every `s01-parse` node, so an attacker who exhausted `MaxAuthTries` produced no `ssh_failed-auth` event at all:

```
error: maximum authentication attempts exceeded for invalid user nonexistentuser123 from 127.0.0.1 port 56876 ssh2 [preauth]
Disconnecting invalid user nonexistentuser123 127.0.0.1 port 56876: Too many authentication failures [preauth]
```

Neither is a near miss of an existing pattern. `SSHD_INVALID_USER` wants a capital `Invalid user` and grok is case sensitive, so the lowercase "for invalid user" in the first line never matches it. `SSHD_PREAUTH_AUTHENTICATING_USER_ALT` wants `Disconnected from ... user X IP`, and the second line is `Disconnecting ... user X IP` with no `from`.

Added `SSHD_MAX_AUTH_ATTEMPTS` and `SSHD_DISCONNECTING_INVALID_USER`, both feeding `ssh_failed-auth` with `target_user`, the same as the neighbouring nodes. The first makes `invalid user ` optional, since OpenSSH omits it when the account actually exists, which is the more interesting case: a real account being brute forced was also going unparsed. There is a third log line in the fixture for that.

Verified with `cscli hubtest`, built from crowdsec master since there is no macOS package. The three lines are appended to `.tests/sshd-logs/sshd-logs.log` and `parser.assert` is regenerated by the tool rather than hand written, so the existing 24 entries come back byte for byte identical and only the two counts and the three new blocks change. All three land as `ssh_failed-auth` with the right `source_ip` and `target_user`.

Reverting just the parser and keeping the fixture fails as it should:

    results["s01-parse"]["crowdsecurity/sshd-logs"][26].Evt.Meta["target_user"] = '""'

The other seven sshd tests are green too, `sshd-logs-fp` among them, which is the one I most wanted to see pass since these patterns are deliberately looser than their neighbours: `sshd-logs`, `sshd-logs-fp`, `sshd-invalid-bf`, `sshd-success-logs`, `sshd-refused-conn`, `sshd-bad-keyexchange-bf`, `sshd_banner_exchange` and `opnsense-sshd`.

### DIFF
--- a/.tests/sshd-logs/parser.assert
+++ b/.tests/sshd-logs/parser.assert
@@ -1,5 +1,5 @@
 len(results) == 3
-len(results["s00-raw"]["crowdsecurity/syslog-logs"]) == 24
+len(results["s00-raw"]["crowdsecurity/syslog-logs"]) == 27
 results["s00-raw"]["crowdsecurity/syslog-logs"][0].Success == true
 results["s00-raw"]["crowdsecurity/syslog-logs"][0].Evt.Parsed["logsource"] == "syslog"
 results["s00-raw"]["crowdsecurity/syslog-logs"][0].Evt.Parsed["message"] == "Invalid user pascal from 35.188.49.176 port 53502"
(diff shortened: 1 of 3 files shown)


arpitjain099:
Withdrawing this one, sorry for the noise. Closing rather than leaving it open for review.

buixor:
Why do you think it's not appropriate ? having a look at the other issue :)

PR #1891, opened 2026-09-11
Apereo cas V7 log support

lctwo:
## Description

A small patch in order to support apereo cas V7 native log format which is not the same at v6 version


sabban:
Hi, 
Thank you for your contribution, could you add some tests for this new format please ?

Thanks,

lctwo:
I'm sorry but i don't really know which test make and how to do it, i'm not a really good developper, so if someone could give assistance.

PR #1892, opened 2026-09-11
Update paperless-ngx-bf.yaml to be consistent and fix leak speed

Jgigantino31:
## Checklist

 - [ ] For new parsers or scenarios, tests have been added 
 - [ ] AI was used to generate any/all content of this PR


sabban:
Hi,

Thank you for your contribution. What is the goal of this change ? What consistency are you aiming for ?

Regards,

Jgigantino31:
Other user enumeration scenarios are labelled as both T1589 and T1110, not just T1589. Add T1110 label to user enumeration scenario.

User enumeration scenario has no impact unless its bucket leaks slower than regular bucket, right now both are 1m. Updated leak speeds to match other similar scenarios on hub, which is 20s for normal scenario and 1m for user enumeration scenario.

No need to reprocess any scenarios here, might have been a mistake since the user enumeration scenario is not reprocessed.

PR #1905, opened 2026-09-23
fix authelia-bf assert for cscli hubtest tiebreaker

blotus:
Sibling of https://github.com/crowdsecurity/crowdsec/pull/4699


sabban (approved):
lgtm

